PR #12, opened 2021-03-22
solve homework requirements using methods, do-while loops and try-parse method.

gregory-butler:
I created methods for each type of user input and BMI calculation. I also made use of do-while loops and try-parse method to not let the user continue without entering correct data. 
LE: Due to curiosity some discussion about data types on the homework channel I researched how to implement a custom number type thus a new class/struct was born: 'SmallNumber'. I replaced number types it in the the BMI calculation with 'SmallNumber'.
Doing so I learned a bit of OOP, what inheritance and interfaces are, how the TryParse pattern works under the hood, exceptions and try-catch statement.


BmiCalculator/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+            Console.WriteLine(welcomeMessage);
+
+            bool isUserContinuing;
+            do
+            {
+                PrintBmi();
+                const string continueMessage = "Calculate BMI for another person? (Y/N)";
+                isUserContinuing = PromptBool(continueMessage, ConsoleKey.Y);
+                Console.WriteLine();
+            } while (isUserContinuing);
```
Almantask, line 24: A bit overkill for this task- but okay.

Almantask (approved):
Highest effort solution so far. Be careful with doing too much, but I appreciate the effort :)

BmiCalculator/Program.cs
mihailfox: Thank you for the review. Sometimes I get carried away when I get excited like I am now about programming with C#. I will try to be more on point for the next homeworks.

PR #15, opened 2021-03-22
Homework1

ASPJulien:


Homework1/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+            Person person = new Person { Name = ConsoleQuestion("full name") };
+            try
+            {
+                person.Age = short.Parse(ConsoleQuestion("age"));
+                person.Weight = double.Parse(ConsoleQuestion("weight (kg)"));
+                person.Height = double.Parse(ConsoleQuestion("height (m)"));
+            }
+            catch (Exception)
+            {
+                Console.WriteLine("One of value that you entered is not correct, please retry");
```
Almantask: Unclear what the actual problem will be

Homework1/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            if (CheckIfValuesCorrect(person))
+            {
+                return person;
+            }
+            Console.WriteLine("One of value that you entered is not correct, please retry");
+            return null;
+        }
+
+        private static string ConsoleQuestion(string valueName)
```
Almantask: What does it mean?

Homework1/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+        private static bool CheckIfValuesCorrect(Person person)
+        {
+            if (person.Age == null || person.Height == null || person.Name == null || person.Weight == null)
+                {
+                    return false;
+                }
+            return true;
+        }
+
+        private static Person GettingInfos()
```
Almantask: UNclear exactly what the info is

Homework1/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+                Console.WriteLine("Do you want to continue? ( y/n )");
+                resp = Console.ReadLine();
+                if (resp == "no" || resp == "n")
+                {
+                    keepGoing = false;
+                }
+            } while (keepGoing);
+        }
+
+        private static bool CheckIfValuesCorrect(Person person)
```
Almantask: What values are correct? Be more specific.

Homework1/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            if (person.Age == null || person.Height == null || person.Name == null || person.Weight == null)
```
Almantask: What if it's empty? Can it be null in the first place if it's user input? Use `string.IsNullOrEmpty()`

Homework1/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+                    }
+                }
+
+                Console.WriteLine("Do you want to continue? ( y/n )");
+                resp = Console.ReadLine();
+                if (resp == "no" || resp == "n")
+                {
+                    keepGoing = false;
+                }
+            } while (keepGoing);
```
Almantask: Too much is going on. Please split into functions.

Almantask (changes requested):

Homework1/Program.cs
ASPJulien: Thanks for the review I will finish when I'll have time

Homework1/Program.cs
ASPJulien: Changed to 
```cs
Console.WriteLine("One or more value that you entered is not a correct number, please retry");
Console.WriteLine("Hint: It could maybe be because you used \".\" instead of \",\"");
```

Homework1/Program.cs
ASPJulien: Changed to
```cs
private static string AskQuestion(string valueName)
```

Homework1/Program.cs
ASPJulien: Changed to
```cs
private static Person GettingPersonInfos()
```

Homework1/Program.cs
ASPJulien: Changed to
```cs
private static bool CheckPersonValuesCorrect(Person person)
```

Homework1/Program.cs
ASPJulien: Is this correct or it will not work ?
```cs
string.IsNullOrEmpty(person.Weight.ToString())
```

Homework1/Program.cs
ASPJulien: Made void AskShowingAllPersons() & bool AskContinue()

ASPJulien:
I've done almost everything you asked, I just have a question for checkIfValuesCorrect

Almantask:
> I've done almost everything you asked, I just have a question for checkIfValuesCorrect

What is the question?

ASPJulien:
How can I check if int is null or empty like the string.isnullorempty ?

Almantask:
> How can I check if int is null or empty like the string.isnullorempty ?

you can if(number == default)

But it just refers to a 0. That's the benefit of numbers or other value types- they are always valid to some extent.

Almantask (approved):

PR #17, opened 2021-03-22
Homework/1

canzoroglu:


BMICalculator/Program.cs
```diff
@@ -0,0 +1,35 @@
...
+                int age = int.Parse(Console.ReadLine());
+                Console.Write("Enter your weight in kg: ");
+                double weight = double.Parse(Console.ReadLine());
+                Console.Write("Enter your height in cm: ");
+                double heightC = double.Parse(Console.ReadLine());
+                double heightM = heightC / 100.0;
+                string msg = $"{name} {surname} is {age} years old, his weight is {weight} kg and his height is {heightC} cm.";
+                double bmi = weight / (heightM * heightM);
+                Console.WriteLine(msg);
+                Console.WriteLine($"{name}'s BMI is {bmi:F2}");
```
Almantask: Split this into functions

BMICalculator/Program.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        static void Main(string[] args)
+        {
+        	// To be able to use "." instead of "," as decimal separator
+            System.Globalization.CultureInfo customCulture = (System.Globalization.CultureInfo)System.Threading.Thread.CurrentThread.CurrentCulture.Clone();
+            customCulture.NumberFormat.NumberDecimalSeparator = ".";
+            System.Threading.Thread.CurrentThread.CurrentCulture = customCulture;
+
+            for (int i = 0; i < 2; i++)
+            {
```
Almantask: A single iteration could be a function call and just call that function twice. Loop is a little overkill.

Almantask (changes requested):

BMICalculator/Program.cs
canzoroglu: Thank you for the review. I've updated the code.

Almantask (approved):
Looks better now. Though it's too early for OOP.

PR #21, opened 2021-03-25
Homework & Recipe app

Svetoslav-Dev:
First github pull

### DIFF
--- a/Homework_1_User_input/Homework_1_User_input/Program.cs
+++ b/Homework_1_User_input/Homework_1_User_input/Program.cs
@@ -1,0 +1,44 @@
+﻿using System;
+
+namespace Homework_1_User_input
+{
+    class Program
+    {
+       
+        static void Main(string[] args)
+        {
+            for (int i = 0; i<=1;) { 
+            Console.WriteLine("Please enter name: ");
+            string name = Console.ReadLine();
+
+            Console.WriteLine("Please enter surname: ");
+            string surname = Console.ReadLine();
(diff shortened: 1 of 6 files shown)


Homework_1_User_input/Homework_1_User_input/Program.cs
```diff
@@ -0,0 +1,44 @@
...
+            Console.WriteLine("Please enter age: ");
+            string ageText = Console.ReadLine();
+            int age = int.Parse(ageText);
+
+            Console.WriteLine("Please enter kg: ");
+            string kgText = Console.ReadLine();
+            double kg = double.Parse(kgText);
+
+            Console.WriteLine("Please enter height: ");
+            string heightText = Console.ReadLine();
```
Almantask, line 26: These two lines could be reuse through a function `Prompt`- which asks to input and reads it.

Homework_1_User_input/Homework_1_User_input/Program.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+
+namespace Homework_1_User_input
+{
+    class Program
+    {
+       
+        static void Main(string[] args)
+        {
+            for (int i = 0; i<=1;) {
```
Almantask, line 10: If instead of a loop you move the body of this function into a function of its own- you won't need a loop and will simply be able to call the function twice.

Almantask (approved):

PR #27, opened 2021-03-28
Homework 1 BMI Calculator Submission

ZHDINC:
Submission for Homework 1. (Apologies for the tuples, hahaha...Really should be a Person class, but getting ahead of ourselves here.)

### DIFF
--- a/LessonOneHomework/Program.cs
+++ b/LessonOneHomework/Program.cs
@@ -1,0 +1,57 @@
+﻿using System;
+
+namespace LessonOneHomework
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            bool continueLoop = true;
+            while(continueLoop)
+            {
+                (string firstName, string surname) = GetUserName();
+                Console.Write("What is your age? ");
+                int age = int.Parse(Console.ReadLine());
+                (double height, double weight) = GetHeightandWeight();
+                double bmi = CalculateBMI(height, weight);
+                Console.WriteLine($"{firstName} {surname} is {age} years old, his weight is {weight} and his height is {height}.");
+                Console.WriteLine($"Their BMI is {bmi}.");
+                Console.Write("Another person? (y/n)");
+                string looper = Console.ReadLine();
+                continueLoop = looper.ToLower() == "n" ? false : true;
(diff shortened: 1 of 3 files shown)


LessonOneHomework/Program.cs
```diff
@@ -0,0 +1,57 @@
...
+                double bmi = CalculateBMI(height, weight);
+                Console.WriteLine($"{firstName} {surname} is {age} years old, his weight is {weight} and his height is {height}.");
+                Console.WriteLine($"Their BMI is {bmi}.");
+                Console.Write("Another person? (y/n)");
+                string looper = Console.ReadLine();
+                continueLoop = looper.ToLower() == "n" ? false : true;
+            }
+        }
+
+        private static (string, string) GetUserName()
```
AlmightyLks, line 25: "Get*" may sound a bit vague, you might as well name it `PromptUsername()` just to be clear about what this does

LessonOneHomework/Program.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            Console.Write("What is your full name? ");
+            string fullName = Console.ReadLine();
+            string[] names = fullName.Split(" ");
+            return (names[0], names[1]);
+        }
+
+        private static (double, double) GetHeightandWeight()
```
AlmightyLks, line 33: Same thing applies here, `Prompt` over `Get`

AlmightyLks (commented):
Overall looking good!

LessonOneHomework/Program.cs
Almantask, line 25: We don't need to have a distinction between a firstname and lastname- no need to return 2 things.

Almantask (approved):
Looks okay, but maybe a little too much overthinking and showcase of advanced features like a tuple.

LessonOneHomework/Program.cs
ZHDINC: Admittedly, it's a job for a class and not tuples, but we haven't covered classes yet so I just did tuples for fun. 

Also, I only made a distinction between firstname and lastname because the requirements listed them distinctly.

PR #29, opened 2021-03-28
Homework Lesson 1 BMI 

QuintenJ:
A little late to the party not sure if I should have skipped this one or not. Figured I would give it a try and get on track.

I attempted to keep it within the scope of the last lesson and not use anything that wasn't covered. It could be more efficient.

### DIFF
--- a/Homework_1/Homework_1/Program.cs
+++ b/Homework_1/Homework_1/Program.cs
@@ -1,0 +1,41 @@
+﻿using System;
+
+namespace Homework_1
+{
+    class Program
+    {
+
+        //Read name, surname, age, weight (in kg) and height(in cm) from console.
+        //Print a message to the screen based on the information given in 1. (an example message is shown below):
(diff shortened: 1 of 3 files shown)


Homework_1/Homework_1/Program.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+
+        //Read name, surname, age, weight (in kg) and height(in cm) from console.
+        //Print a message to the screen based on the information given in 1. (an example message is shown below):
+        //Tom Jefferson is 19 years old, his weight is 50 kg and his height is 156.5 cm.
+        //Calculate and print body-mass index(BMI)
+        //Do 1 and 2 for another person.
+        static void Main(string[] args)
+        {
+            DisplayToConsole();
```
Almantask, line 15: Does the method name tell what it does? - I don't think so.
Rename the method so that it says what it does.

Almantask (approved):
It's okay what you did, even though I was expecting more functions- what you have works too and is not worse than others.

PR #31, opened 2021-03-29
implement array operations methods and simple login system

gregory-butler:
I implemented methods for each array operation and stored them in their own static class named 'Arrays'. For each of the method in the 'Arrays' class I created an correspondent in the 'Demo' class to demonstrate and test that the methods work correctly.
While I was working on this I discovered how to create XML comments and wrote them for the 'Arrays' class.
For the last part of the homework I created another project called 'LoginSimulator' in which I implemented two methods: one to prompt user for the name and password without showing pressed keys and the other to test credential against stored one.
LE:
I updated the logic for the 'Sort' method to use 'Merge Sort' algorithm, before I used 'Bubble Sort'. Like with the other methods I found the need to test it with different data and in different scenarios so I decided to delve a bit in to unit testing. 
The framework I chose is xUnit. 
LELE: found out that triple A is not just for battery's 😄


ArrayOperationTests/InsertAtTests.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Runtime.InteropServices;
+using Xunit;
+using ArrayOperations;
+
+namespace ArrayOperationTests
+{
+  public class InsertAtTests
+  {
+    [Fact]
+    public void InsertAt_Should_Throw_ArgumentException_When_Input_Array_Is_Empty()
```
westonpace: Why?  Why can't I insert into an empty array at index 0?

ArrayOperationTests/InsertAtTests.cs
```diff
@@ -0,0 +1,49 @@
...
+    [InlineData(new[] {0,2,3}, 1, 1, new[] {0, 1, 2, 3})]
+    [InlineData(new[] {1,2,3}, 0, 0, new[] {0, 1, 2, 3})]
+    [InlineData(new[] {0,1,2}, 3, 3, new[] {0, 1, 2, 3})]
+    [InlineData(new[] {0},1,1, new[] {0,1})]
+    [InlineData(new[] {1},0,0, new[] {0,1})]
+    public void InsertAt_Should_Return_New_Array_With_Item_Inserted_At_Correct_Position(int[] input, int number, int index, int[] expected)
+    { 
+      int[] actual; 
+       
+      actual = Arrays.InsertAt(input, number, index);
```
westonpace: You could combine the declaration and the call here...
`int[] actual = Arrays.InsertAt(input, number, index);`

ArrayOperationTests/SortTests.cs
```diff
@@ -0,0 +1,57 @@
...
+  public class SortTests
+  {
+    [Fact]
+    public void Sort_Should_Throw_ArgumentException_When_Input_Array_Is_Empty()
+    {
+      var input = new int[] {};
+      
+      Action actual = () => Arrays.Sort(input);
+      
+      Assert.Throws<ArgumentException>(actual);
```
westonpace: Again, why?  Isn't an empty array sorted?

ArrayOperationTests/SortTests.cs
```diff
@@ -0,0 +1,57 @@
...
+      Assert.Equal(input,actual);
+    }
+
+    [Theory]
+    [InlineData(new[] {1, 0}, new[] {0, 1})]
+    [InlineData(new[] {0, 1}, new[] {0, 1})]
+    [InlineData(new[] {1, 0, 2, -1, 3}, new[] {-1, 0, 1, 2, 3})]
+    [InlineData(new[] {1, -1, 1}, new[] {-1, 1, 1})]
+    [InlineData(new[] {1, 1, 1}, new[] {1, 1, 1})]
+    public void Sort_Should_Return_Expected_Array(int[] input, int[] expected)
```
westonpace, line 28: Not a review comment so much but something to make you think.

What you're doing here is very common.  You supply the input and the expected output.

Another way to do this kind of testing is to supply an input and a verification algorithm.  For example, for a sorting algorithm you can verify by taking the result and making sure each item is >= the previous.  Then you don't have to supply the expected output.

Anyways, what you have is probably best.  I'm just rambling.

ArrayOperations/Arrays.cs
```diff
@@ -0,0 +1,324 @@
...
+{
+  /// <summary>
+  /// This class contain methods for array operations.
+  /// </summary>
+  public static class Arrays
+  {
+    /// <summary>
+    /// Sort an array using 'Bubble Sort' algorithm.
+    /// </summary>
+    /// <param name="input">Source array.</param>
```
westonpace: If you're going to do this commenting (you don't need to) then you should also include `<returns>`

ArrayOperations/Arrays.cs
```diff
@@ -0,0 +1,324 @@
...
+
+namespace ArrayOperations
+{
+  /// <summary>
+  /// This class contain methods for array operations.
+  /// </summary>
+  public static class Arrays
+  {
+    /// <summary>
+    /// Sort an array using 'Bubble Sort' algorithm.
```
westonpace: I'm pretty sure you're not using bubble sort anymore.

ArrayOperations/Arrays.cs
```diff
@@ -0,0 +1,324 @@
...
+
+      if (input.Length == 1)
+      {
+        return  input;
+      }
+
+      var midPoint = input.Length / 2;  
+      
+      var left = new int[midPoint];
+      var right = input.Length % 2 == 0 ? new int[midPoint] : new int[midPoint + 1];
```
westonpace, line 25: This is great and it utilizes what was taught in class more effectively.  I'm only bringing this up since you are clearly putting in a lot of effort here.  For a merge sort on very large arrays, it's ideal not to have to do all these new arrays and copies.  Making changes in-place will have better performance.  If you want to make one copy at the top-level you can do that too and then you preserve the input array which is often good.

ArrayOperations/Arrays.cs
```diff
@@ -0,0 +1,324 @@
...
+            indexOutput++;
+          }
+        }
+        else if (indexLeft < left.Length)
+        {
+          output[indexOutput] = left[indexLeft];
+          indexLeft++;
+          indexOutput++;
+        }
+        else if (indexRight < right.Length)
```
westonpace: Technically this could just be `else` I think.

ArrayOperations/Arrays.cs
```diff
@@ -0,0 +1,324 @@
...
+          indexOutput++;
+        }  
+      }
+      return output;
+    }
+ 
+    /// <summary>
+    /// Inserts a new element in an array at the specified position.
+    /// </summary>
+    /// <param name="input">Destination array.</param>
```
westonpace: I wouldn't use the word `Destination`.  The input array is going to be unmodified.  Instead you are returning a new array.

ArrayOperations/Arrays.cs
```diff
@@ -0,0 +1,324 @@
...
+    /// Inserts a new element at the beginning of the array.
+    /// </summary>
+    /// <param name="input">Destination array.</param>
+    /// <param name="number">The number to be inserted.</param>
+    /// <returns>A new array with it's size increased by one and a new element located at the front of the array.</returns>
+    /// <exception cref="ArgumentNullException">The source array is not initialized.</exception>
+    /// <exception cref="ArgumentException">The destination array has no elements.</exception>
+    public static int[] InsertFirst(int[] input, int number)
+    {
+      if (input == null)
```
westonpace: These checks are redundant.  You will catch this in `InsertAt`.

ArrayOperations/Arrays.cs
```diff
@@ -0,0 +1,324 @@
...
+        throw new ArgumentNullException(nameof(input));
+      }
+
+      if (input.Length == 0)
+      {
+        throw new ArgumentException("Value cannot be an empty collection.", nameof(input));
+      }
+      
+      const int index = 0;
+      var output = InsertAt(input, number, index);
```
westonpace: Don't create variables just to return them.  Do `return InsertAt(input, number, index);` instead.

ArrayOperations/Arrays.cs
```diff
@@ -0,0 +1,324 @@
...
+        output[i] = i < index ? input[i] : input[i + 1];
+      }
+
+      return output;
+    }
+    
+    /// <summary>
+    /// Removes the first element from the array.
+    /// </summary>
+    /// <param name="input">Source array.</param>
```
westonpace: Minor nit: Now you're calling it source array which also doesn't make sense.  I get there is some difference "I am removing an element from...source" vs "I am adding an element to...destination".  But still, I think I'd just say "The array to remove an element from" or something like that.  I'm not sure source or destination are the right terms.

ArrayOperations/Arrays.cs
```diff
@@ -0,0 +1,324 @@
...
+      {
+        throw new ArgumentException("Value cannot be an empty collection.", nameof(input));
+      }
+      
+      if (string.IsNullOrWhiteSpace(message))
+      {
+        throw new ArgumentException("Value cannot be null or empty.", nameof(message));
+      }
+
+      Console.WriteLine(message);
```
westonpace: Although it does shorten things I think having `message` here violates the "single responsibility principle" (more on that later) and so I'd keep...
```
Console.WriteLine(message);
Print(arr);
```
over trying to combine the two.

ArrayOperations/Demo.cs
```diff
@@ -0,0 +1,111 @@
...
+      Arrays.Print(input, "Initial array elements:");
+      var output = Arrays.Sort(input);
+      Arrays.Print(output, "Array after sort operation:");
+    }
+
+    private static void InsertAnElementAtBeginningOfTheArray(int[] input)
+    {
+      Arrays.Print(input,"Initial array elements:");
+
+      const int numberToInsert = 12;
```
westonpace: Use `NumberToInsert` (first letter capitalized) for constants.

LoginSimulator/Program.cs
```diff
@@ -0,0 +1,56 @@
...
+                              string.Equals(userPassword, storedUserPassword);
+      if (isValidCredential)
+      {
+        Console.WriteLine(successfulLoginMessage);
+      }
+    }
+    
+    private static string PromptString(string message, bool isKeyVisible = false)
+    {
+      if (string.IsNullOrWhiteSpace(message))
```
westonpace, line 32: In general, don't do null checks for internal methods.  You will drive yourself crazy.

LoginSimulator/Program.cs
```diff
@@ -0,0 +1,56 @@
...
+      {
+        throw new ArgumentException("Value cannot be null or empty.", nameof(message));
+      }
+      
+      Console.Write(message);
+      
+      var output = "";
+      while (true)
+      {
+        var pressedKey = Console.ReadKey(!isKeyVisible);
```
westonpace, line 42: Why not `Console.ReadLine`?

westonpace (commented):
Clearly you put a lot of work into this.  So I decided to be extra strict.  It looks great.

LoginSimulator/Program.cs
gregory-butler: I chose to use  `Console.ReadKey` so I can hide the user input. From my findings this is the only way I could achieve this.

ArrayOperationTests/InsertAtTests.cs
gregory-butler: I only split the initialization from declaration after I read about `Arrange, Act Assert`. I can see now the error in my logic as the 'Arrange' part is done using `InlineData`.

ArrayOperationTests/InsertAtTests.cs
gregory-butler: Makes a lot of sense when you are asking, the user should be able to insert an item in an empty array.

ArrayOperationTests/SortTests.cs
gregory-butler: It's definitely not easy to read. I don't like how the inline data looks, even more so for the other tests that have more arguments. I saw there other ways to do it but it requires more OOP knowledge than I currently poses. One thing I've learned is that it's not a good idea to introduce logic in the actual tests.

ArrayOperations/Arrays.cs
gregory-butler: I just forgot to add it. I initially implemented bubble sort to do the sorting in place without returning another array.

ArrayOperations/Demo.cs
gregory-butler: By default Visual Studio and Rider recommend to use camel case and it stuck with me.

ArrayOperations/Arrays.cs
gregory-butler: I definitely want to become as experienced as you are to notice these small things. Thank you 🤗

ArrayOperations/Arrays.cs
gregory-butler: I understand, definitely something to study in the following days.

ArrayOperations/Arrays.cs
gregory-butler: Writing comments is cumbersome but I think it's good practice to describe what I was trying to achieve. Thank you.

ArrayOperations/Arrays.cs
gregory-butler: I understand. Thank you.
Does the 'PromptString' method from the 'LoginSimulator' violate SRP?

gregory-butler:
Thank you very much for the detailed review. I really appreciate the feedback you have given me.
I updated my code and I will push the commits shortly.

ArrayOperations/Arrays.cs
westonpace: It's not always black&white.  I think PromptString is fine.

LoginSimulator/Program.cs
westonpace: That's a cool trick and makes a lot of sense.

ArrayOperations/Arrays.cs
westonpace: In general people don't comment every method they write.  Usually they separate their project (mentally or otherwise) into a "public" portion, methods that are meant to be called by other people that use your code, and a private portion, helper methods that help implement the public methods.  For every 1 public method (that you have to comment) you might have 10-50 private methods (that people usually don't comment, or they do so only briefly).

PR #33, opened 2021-03-31
Homework/2

cjahermanns:


Homework2.Tests/ArrayExtensionsTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Homework2.Tests
+{
+    public class ArrayExtensionsTests
+    {
+        [Theory]
+        [InlineData(999, 4)]
+        [InlineData(999, 0)]
+        [InlineData(999, 10)]
+        public void AddAt_adds_at_idx(int item, int idx)
```
westonpace, line 12: The assignment didn't ask for bounds checking (or even testing) so you're already above and beyond.  If you were going for real world application you could add a test for adding an item to an empty array or a null array, or an empty index or an index beyond the length of the data.

Homework2.Tests/ArrayExtensionsTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            Assert.Equal(testarray.Length - 1, newarray.Length);
+            Assert.DoesNotContain(testarray[idx], newarray);
+        }
+
+        [Theory]
+        [InlineData(4, 5, 6, 1, 3, 1, 3, 4, 5, 6)]
+        [InlineData(0, 1, 2, 3, 9, -5, -5, 0, 1, 2, 3, 9)]
+        [InlineData(9.1, 8.5, 1.0, -5.2, -5.2, 1.0, 8.5, 9.1)]
+        public void Sort_sorts_the_array(params object[] objs)
```
westonpace, line 40: While this method of testing is clever it is a little confusing.  Why not just take in two arguments, the original array and the sorted array.  This way you don't have to mess with slicing and dicing the array in the test.  Alternatively, just take in the unsorted array and then verify the result by making sure each item is >= the item that came before.

Homework2/ArrayExtension.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            T[] output = new T[newSize];
+
+            for (int i = 0; i < newSize; i++)
+            {
+                if (i == idx)
+                {
+                    output[i] = item;
+
+                    for (int j = idx; j < array.Length; j++)
```
westonpace, line 19: Minor nit: This works but it is maybe not the most intuitive approach to solving this problem.  You could do this all with one loop or with two loops one after the other.  It works so I'm not complaining, just observing.

Homework2/ArrayExtension.cs
```diff
@@ -0,0 +1,96 @@
...
+
+                output[i] = array[i];
+            }
+
+            return output;
+        }
+
+        public static T[] RemoveAtBeginning<T>(this T[] array)
+        {
+            return array.RemoveAt(0);
```
westonpace, line 71: Good work using your general purpose method to implement the specialized methods.

Homework2/ArrayExtension.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        public static T[] RemoveAtEnd<T>(this T[] array)
+        {
+            return array.RemoveAt(array.Length - 1);
+        }
+
+        public static void Sort<T>(this T[] array)
+        {
+            //var arraycopy = new T[array.Length];
```
westonpace, line 81: Minor nit: Clean up dead code before submitting anything.

Homework2/ArrayExtension.cs
```diff
@@ -0,0 +1,96 @@
...
+            //array.CopyTo(arraycopy, 0);
+
+            var arraycopy = array.ToList();
+
+            for (int i = 0; i < array.Length; i++)
```
westonpace, line 86: This works but is not the most efficient approach.  We have not spoken about efficiency yet (and premature optimization is the root of all evil). However...

You first make a copy of the array, but you are modifying the array in-place so this is only a convenience copy and not strictly neccesary.

Every call to Min is going to loop through the entire array.  So this puts it on par with bubble sort / insertion sort for effeciency.

Every call to Remove is going to loop through the array again.

Potential optimizations (not required for the class):

* Use a more efficient algorithm
* Grab the min index instead of just the min and then use that to remove
* Don't make a copy of the array and modify in-place or make the copy useful and return it.

Homework2/Program.cs
```diff
@@ -0,0 +1,28 @@
...
+            do
+            {
+                Console.WriteLine("Please enter your username:");
+                var user = Console.ReadLine();
+                Console.WriteLine("Please enter your password:");
+                var pw = Console.ReadLine();
+
+                loginSucceeded =
+                    string.Equals(user, username, StringComparison.OrdinalIgnoreCase) &&
+                    string.Equals(pw, password, StringComparison.Ordinal);
```
westonpace, line 22: `StringComparison.Ordinal` is the default, you don't actually need to specify it here although it doesn't hurt.

Homework2/Program.cs
```diff
@@ -0,0 +1,28 @@
...
+            } while (!loginSucceeded);
```
westonpace, line 23: If you're going to loop in this way it is generally nice to show some kind of "username or password incorrect, please try again" message.

westonpace (commented):
Very good.  Your code meets and goes beyond the requirements.  I added a few comments.

Homework2.Tests/ArrayExtensionsTests.cs
cjahermanns: Those tests definitely should be there. Since I did my submission late at night I tried to stick with the happy path tests though... Maybe in future submissions my tests will be much better :D

Homework2.Tests/ArrayExtensionsTests.cs
cjahermanns: That's so true! Don't know why I didn't just use two arrays instead of their elements...

Homework2/ArrayExtension.cs
cjahermanns: Tyvm, I try writing code with SOLID and DRY in mind.

Homework2/ArrayExtension.cs
cjahermanns: My bad will keep that in mind!

Homework2/ArrayExtension.cs
cjahermanns: Actually wanted to do some refactoring in the last couple of days but did not find the time to do so :D

Homework2/ArrayExtension.cs
cjahermanns: I guess you're right... Those LINQ queries just are so convenient :D

Homework2/Program.cs
cjahermanns: Wasn't sure about that so I went with the additional argument :D

Homework2/Program.cs
cjahermanns: You're right, that would be much better!

Almantask (approved):

PR #34, opened 2021-04-01
Second homework

unbekanntunity:


.vs/ProjectSettings.json
```diff
@@ -0,0 +1,3 @@
+{
```
westonpace: Unless you have a good reason the `.vs` folder shouldn't be checked into source control.

BootCampV2/Homeworks/SecondWeek/ArrayExtension.cs
```diff
@@ -0,0 +1,149 @@
+using System;
+
+namespace BootCampV2
+{
+    public class ArrayExtension
```
westonpace: C# has a special feature called "Extension methods" that we haven't spoken about yet in class.  In general, files whose name ends in "Extensions" or "Extension" are adding extension methods.  Since you don't do that here you shouldn't name this file `ArrayExtension`.

BootCampV2/Core.cs
```diff
@@ -0,0 +1,15 @@
+﻿using BootCampV2.Homeworks.SecondWeek;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCampV2
+{
+    class Core
+    {
+        static public void Main(string[] args)
```
westonpace: I'm not sure what is going on here.

BootCampV2/Homeworks/SecondWeek/ArrayExtension.cs
```diff
@@ -0,0 +1,149 @@
...
+    {
+        /// <summary>
+        /// Use the bubble sort algorithm to sort an array
+        /// <para> return: sorted Array </para> 
+        /// </summary>
+        public static int[] BubbleSort(int[] array)
+        {
+            for (int i = 0; i < array.Length; i++)
```
westonpace: This is a good implementation of bubble sort.

* Your inner loop doesn't loop the entire array.
* You bail early if you find the rest of the array is sorted
* Nested `SwapElements` method

Good work.

BootCampV2/Homeworks/SecondWeek/ArrayExtension.cs
```diff
@@ -0,0 +1,149 @@
...
+                    if (array[m] > array[m + 1])
+                    {
+                        array = SwapElements(array, m, m + 1);
+                        swap = true;
+                    }
+                }
+                if (!swap)
+                    break;
+            }
+            return array;
```
westonpace: Typically I would expect a sort method to either modify the array I pass in (in which case the return type is void) or make a copy of the array and return the sorted array as a copy.  Here you are returning an array (so I might think you are making a copy) but you modify the array passed in.  This would be a little confusing to a user.

BootCampV2/Homeworks/SecondWeek/ArrayExtension.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+
+        /// <summary>
+        /// Swap two elements with each other
+        /// <para> return: modified Array </para> 
+        /// </summary>
+        public static int[] SwapElements(int[] array, int firstIndex, int secondIndex)
+        {
+            if (firstIndex >= array.Length || firstIndex < 0 ||
```
westonpace: Normally I might think SwapElements to be a private helper method, then you wouldn't have to do validation (if you pass bad input to a helper method you can generally just let it throw an exception, input validation should happen when input comes from external sources) and you wouldn't have to comment it (typically you don't comment private helper methods but only methods you are making for other people to use).  By making it public you are making more work for yourself.  It isn't a bad thing necessarily but something to consider.

BootCampV2/Homeworks/SecondWeek/ArrayExtension.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        /// <summary>
+        /// Insert an element at the given position in the array
+        /// <para> return: modified Array </para> 
+        /// </summary>
+
+        public static T[] InsertAt<T>(T[] array, T element, int index)
+        {
+            if (index >= array.Length || index < 0)
```
westonpace: If you're going to do input validation you should also check to make sure `array` is not `null`.

BootCampV2/Homeworks/SecondWeek/ArrayExtension.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        /// <summary>
+        /// Remove the first element of the array
+        /// <para> return: modified Array </para> 
+        /// </summary>
+
+        public static T[] RemoveFirst<T>(T[] array)
+        {
+            if (array.Length < 1)
```
westonpace: You could use your general purpose `RemoveAt` to implement `RemoveFirst` and `RemoveLast`

BootCampV2/Homeworks/SecondWeek/ArrayExtension.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            return expandedArray;
+        }
+
+        /// <summary>
+        /// Remove an element of the given index of the array
+        /// <para> return: modified Array </para> 
+        /// </summary>
+
+        public static T[] RemoveAt<T>(T[] array, int index)
```
westonpace: Good work using generics.  What would be the challenges involved with creating a generic sort?

BootCampV2/Homeworks/SecondWeek/Login.cs
```diff
@@ -0,0 +1,54 @@
...
+            if(validUsername && validPassword)
+                Console.WriteLine("Logged In");
+            else
+                Console.WriteLine("Wrong username or password");
+        }
+
+        public bool ValidateUsername(string username)
+        {
+            var adaptedUsername = Username.ToLower();
+            var adaptedusername = username.ToLower();
```
westonpace: I'm not a huge fan of these variables differing only by case.

BootCampV2/Homeworks/SecondWeek/Login.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            if (adaptedUsername.Equals(adaptedusername))
```
westonpace: Could simplify to...

`return adaptedUsername.Equals(adaptedusername);`

BootCampV2/Homeworks/SecondWeek/Login.cs
```diff
@@ -0,0 +1,54 @@
...
+                return true;
+
+            return false;
+        }
+
+        public bool ValidatePassword(string password)
+        {
+            if (Password.Equals(password))
```
westonpace: Could simplify to `return Password.Equals(password);`

BootCampV2/Homeworks/SecondWeek/Login.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Text;
+
+namespace BootCampV2.Homeworks.SecondWeek
+{
+    class Login
+    {
+        public string Username{ get; private set; }
+        public string Password{ get; private set; }
+
+        public void SetData(string username, string password)
```
westonpace: What happens if you forget to call `SetData`?  Better to pass these values in via the constructor instead of a separate method and then you don't have to worry about this possibility.  Of course, we haven't covered OOP yet :)

BootCampV2/bin/Debug/netcoreapp3.1/BootCampV2.deps.json
```diff
@@ -0,0 +1,23 @@
+{
```
westonpace, line 1: Definitely don't checkin your `bin` folder.

westonpace (commented):
Looks great.  You satisfied all the requirements and I can see you went above and beyond using generics, classes, etc.

Almantask:
Please cleanup and resolve the remarks before I can mark this as done.

unbekanntunity:
ty

PR #35, opened 2021-04-01
Homework/2

AyameU:
Hello, here's my homework for week 2. I kept things pretty basic and tried to just use loops and if statements. Hopefully my methods are organized better this time. :)

Now that I don't have have to stay within Discord's character limit, I added documentation to the methods as well.

### DIFF
--- a/ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
+++ b/ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
@@ -1,0 +1,105 @@
+﻿using System;
+
+namespace ControlFlowArraysAndStrings
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            string[] fruits = { "banana",
+                                "apple", 
+                                "orange", 
+                                "pineapple", 
+                                "blueberry",
+                                "raspberry",
+                                "apricot" };
+
+            string[] sortedFruits = SortArray(fruits);
+
+            Console.Write("Unsorted: ");
+            PrintArray(fruits);
+
+            Console.Write("Sorted: ");
+            PrintArray(sortedFruits);
+        }
+
+        /// <summary>
+        /// Sorts a copy of the string array in alphabetical order.
+        /// </summary>
+        /// <param name="array">The array to be sorted.</param>
+        /// <returns>A sorted copy of the specified array.</returns>
+        public static string[] SortArray(string[] array)
+        {
+            string[] sorted = CopyStringArray(array);
+
+            for (int i = 0; i < sorted.Length; i++)
+            {
+                bool isShifted = false;
+
+                for (int j = 0; j < sorted.Length; j++)
+                {
+                    // Check that this is not the last index in the array.
+                    if (j != sorted.Length - 1)
+                    {
+                        string shiftedValue;
+
+                        // Check if CompareTo() return 1. 
+                        // If it does, switch positions.
+                        if (sorted[j].CompareTo(sorted[j + 1]) == 1)
+                        {
+                            shiftedValue = sorted[j];
+
+                            sorted[j] = sorted[j + 1];
+
+                            sorted[j + 1] = shiftedValue;
+
+                            isShifted = true;
+                        }
+                    }             
+                }
+
+                // If no value has been changed after one full pass 
+                // of the inner loop, break out of the outer loop.
+                if (!isShifted)
+                {
+                    break;
+                }
+            }
+
+            return sorted;
+        }
+
+        /// <summary>
+        /// Copies the specified array and returns a new array.
+        /// </summary>
+        /// <param name="array">The array to be copied.</param>
+        /// <returns>A copy of the specified array.</returns>
+        public static string[] CopyStringArray(string[] array)
+        {
+            string[] newArray = new string[array.Length];
+
+            Array.Copy(array, newArray, array.Length);
+
+            return newArray;
+        }
+
+        /// <summary>
+        /// Prints the array to the console.
+        /// </summary>
+        /// <param name="array">The array to be printed.</param>
+        public static void PrintArray(string[] array)
+        {
+            for (int i = 0; i < array.Length; i++)
+            {
+                if (i != array.Length - 1)
+                {
+                    Console.Write(array[i] + ", ");
+                }
+                else
+                {
+                    Console.Write(array[i] + "\n");
+                }
+            }
+        }
+    }
+}
--- a/ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
+++ b/ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
@@ -14,7 +14,8 @@ namespace ControlFlowArraysAndStrings
                                 "raspberry",
                                 "apricot" };
 
-            // Testing the Sort method.
+            // Sort method.
+            Console.WriteLine("\nTesting the Sort method:");
             string[] sortedFruits = Sort(fruits);
 
             Console.Write("Unsorted: ");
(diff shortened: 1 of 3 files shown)


ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
```diff
@@ -0,0 +1,343 @@
...
+        public static string[] Sort(string[] array)
+        {
+            string[] sorted = Copy(array);
+
+            for (int i = 0; i < sorted.Length - 1; i++)
+            {
+                for (int j = 0; j < sorted.Length - i - 1; j++)
+                {
+                    // Check if CompareTo() returns 1, if it does, swap positions.
+                    if (sorted[j].CompareTo(sorted[j + 1]) == 1)
```
westonpace: The fact that `String.CompareTo` returns `1` or `-1` is an implementation detail.  According to the docs for the method (https://docs.microsoft.com/en-us/dotnet/api/system.string.compareto?view=net-5.0)...

```
Less than zero | This instance precedes strB.
Zero | This instance has the same position in the sort order as strB.
Greater than zero | This instance follows strB.
```

So technically it is allowed to return any number greater than 0 in this case.  You should test for that `CompareTo(...) > 0` and not `CompareTo(...) == 1` to prevent possibly getting bugs when you upgrade from one .NET version to the next.

ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
```diff
@@ -0,0 +1,343 @@
...
+        {
+            for (int i = 0; i < array.Length; i++)
+            {
+                if (i != array.Length - 1)
+                {
+                    Console.Write($"{array[i]}, ");
+                }
+                else
+                {
+                    Console.Write($"{array[i]}\n");
```
westonpace: Regrettably, some windows tools will not recognize `\n` properly.  (https://stackoverflow.com/questions/3986093/in-c-whats-the-difference-between-n-and-r-n).  It's best to use `WriteLine` (and drop the `\n`) or use `Environment.NewLine`.

ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
```diff
@@ -0,0 +1,343 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Adds a specified value to the beginning of the specified array
```
westonpace, line 136: Nit: drop "specified".  Having it twice here feels a little redundant.  Also, you're not technically adding the value to the "specified" array because you are returning a copy.

ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
```diff
@@ -0,0 +1,343 @@
...
+        }
+
+        /// <summary>
+        /// Removes the value at the end of the specified array.
+        /// </summary>
+        /// <param name="array">The array to be changed.</param>
+        /// <returns>A new array with a value removed from the end.</returns>
+        public static string[] RemoveElementFromEnd(string[] array)
+        {
+            string[] newArray = new string[array.Length - 1];
```
westonpace: You could use your general purpose `RemoveElementFromAnywhere` to implement the specialized `RemoveElementFromEnd`.

ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
```diff
@@ -0,0 +1,343 @@
...
+            }
+
+            return newArray;
+        }
+
+        /// <summary>
+        /// Removes the value at the specified index from the specified array.
+        /// </summary>
+        /// <param name="array">The array to be changed.</param>
+        /// <param name="index">The index to be removed.</param>
```
westonpace: Nit: `The index to be removed` -> `The index of the value to be removed`.

ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
```diff
@@ -0,0 +1,343 @@
...
+            {
+                while (!isValid[i] && loginAttempts != 0)
+                {
+                    string credential = GetLoginCredentialsFromUser(credentialType[i]);
+                    isValid[i] = ValidateCredential(credentialType[i], credential);
+
+                    if (!isValid[i])
+                    {
+                        loginAttempts--;
+                        Console.WriteLine($"Invalid {credentialType[i]}. {loginAttempts} tries left.");
```
westonpace, line 257: I'm not running this to test but if I try entering a bad username it looks like it might still prompt me for a password?  Also, not really part of the assignment, but if you tell me as soon as I enter the username then it is technically a security issue because you give away which usernames are valid and which are not.

ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
```diff
@@ -0,0 +1,343 @@
...
+        }
+
+        /// <summary>
+        /// Attempts to log in a user and prints a message to the console if login 
+        /// was successful or unsuccessful. Allows 3 unsuccessful attempts before lock out.
+        /// </summary>
+        public static void Login()
+        {
+            int loginAttempts = 3;
+            string[] credentialType = { "user name", "password"};
```
westonpace, line 244: This is a good attempt at generalizing (both "username" and "password" are "credentials" and so you can write logic to handle arbitrary credentials and you don't have to write two sets of similar logic) but it falls short.  You end up with some confusing code and then you have that awkward if statement in `ValidateCredential` (this sort thing is called a "leaky abstraction").

Later, when we learn OOP, or more functional concepts, there are ways you could maybe achieve the goal you set out for.

ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
```diff
@@ -0,0 +1,343 @@
...
+            bool isValid = false;
+
+            if (credentialType.Equals("user name", StringComparison.OrdinalIgnoreCase))
+            {
+                string hiddenUsername = "CSharp";
+                isValid = credential.Equals(hiddenUsername,
+                                            StringComparison.OrdinalIgnoreCase);
+            }
+
+            if (credentialType.Equals("password", StringComparison.OrdinalIgnoreCase))
```
westonpace: This is not correct.  The requirements ask you to compare passwords case sensitively.

westonpace (commented):
Good submission.  I appreciate the concise and sufficient demo in `Main`.  Also, you are consistent with your array manipulation methods in that they all return a copy of the array and do not modify the array passed in (either approach would have been fine but consistency is a bonus).  I've added a few comments.

ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
AyameU: Oh interesting, thank you for pointing that out.

ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
AyameU: That's good to know. I use `\n` a lot. I'll switch to `Environment.NewLine` from now on.

ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
AyameU: It prompts for a username until a valid one is entered or `loginAttempts` hits 0. If the attempts are at 0, it'll break out of the outer loop and "lock" them out, they won't get prompted for a password. If they haven't run out of attempts and the username is valid it will prompt for a password.

ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
AyameU: Wow that suggestion really simplified the `RemoveElementFromEnd`. Thanks!

ControlFlowArraysAndStrings/ControlFlowArraysAndStrings/Program.cs
```diff
@@ -0,0 +1,311 @@
...
+
+            string[] removeStart = RemoveElementFromAnywhere(fruits, 0);
+            Print(removeStart);
+
+            string[] removeEnd = RemoveElementFromAnywhere(fruits, 6);
+            Print(removeEnd);
+
+            // Login method.
+            Console.WriteLine($"{Environment.NewLine}Testing the Login method:");
+            Login();
```
Almantask, line 70: Instead of those code blocks with comments, you could have the same done in methods and then wouldn't need comments.

instead of:

```cs
            // Sort method.
            Console.WriteLine($"{Environment.NewLine}Testing the Sort method:");
            string[] sortedFruits = Sort(fruits);
            Print(sortedFruits);
```

```cs
            void DemoSortMethod()
            {
                Console.WriteLine($"{Environment.NewLine}Testing the Sort method:");
                string[] sortedFruits = Sort(fruits);
                Print(sortedFruits);
            }
```

Almantask (approved):
Looks great!

PR #38, opened 2021-04-03
Homework/2-3

Laimonukas:
Login, adding/removing/sorting users.

### DIFF
--- a/Homework/2/Program.cs
+++ b/Homework/2/Program.cs
@@ -1,0 +1,390 @@
+﻿using System;
+
+namespace Homework2
+{
+    class Program
+    {
+        //Create a method for each operation:
+        //Sort an array
+        //Add element at the start of an array
+        //Add element at the end of an array
+        //Add element at any position of an array
+        //Remove element at the start of an array
+        //Remove element at the end of an array
+        //Remove element at a given index of an array
+        //Login: takes username and password.If they match a hidden username and password- a message "Logged in!" is printed.
+        //Username capitalization does not matter, password capitalization must match.
+
+        //Variables
+        private static bool loggedIn = false; // Variable to track if user is "logged in"
+        private static string[][] userArray = new string[5][]; // array of user arrays with login info which is string[username,password]; Would rather use List<>
+        
+
+        static void Main(string[] args)
+        {
+            //Adding a couple of users to array
+            userArray[0] = new string[] { "Admin", "slAptazodis" };
+            userArray[1] = new string[] { "user1", "pSwD" };
+            userArray[2] = new string[] { "LaIMONAS", "123" };
+            userArray[3] = new string[] { "Xavier", "Dust" };
+            userArray[4] = new string[] { "Blaber", "Lol" };
+            MenuSelection();
+        }
+
+        //Main function which lists available options
+        static void MenuSelection() {
+            Console.Clear();
+
+            if (loggedIn)
+            {
+                Console.WriteLine("Logged in. \n\n");
+                Console.WriteLine("3 - List users\n");
+                Console.WriteLine("4 - Add user to start of list\n");
+                Console.WriteLine("5 - Add user to end of list\n");
+                Console.WriteLine("6 - Add user to n'th position in list\n");
+                Console.WriteLine("7 - Remove user from start of list\n");
+                Console.WriteLine("8 - Remove user from end of list\n");
+                Console.WriteLine("9 - Remove n'th user\n");
+                Console.WriteLine("10 - Sort users by username (asc/dec)\n");
+                Console.WriteLine("11 - Log out\n");
+            }
+            else {
+                Console.WriteLine("Select option:\n");
+                Console.WriteLine("1 - Log in\n");
+                Console.WriteLine("2 - Register\n");
+            }
+
+            OptionSelection(option: Console.ReadLine());
+        }
+
+        //Funtion that gets which option was selected and calls other needed functions via switch statement.
+        static void OptionSelection(string option) {
+            Console.Clear();
+            switch (option)
+            {
+                case "1":
+                    // Log in
+                    Login();
+                    break;
+                case "2":
+                    //Register / adding to end of list
+                    PromptAddUserToArray();
+                    break;
+                case "3":
+                    //Listing all users;
+                    ListUsers();
+                    break;
+                case "4":
+                    //Adding to start of list
+                    PromptAddUserToArray(addToEnd: false, addToStart: true);
+                    break;
+                case "5":
+                    //adding to end of list/ like register
+                    PromptAddUserToArray();
+                    break;
+                case "6":
+                    //Adding to n'th(index) position on list
+                    PromptAddUserToArray(addToEnd: false, addToStart: false);
+                    break;
+                case "7":
+                    //Remove from start of the array
+                    PromptRemoveUser(removeFromEnd: false, removeFromStart: true);
+                    break;
+                case "8":
+                    //Remove from end of the array
+                    PromptRemoveUser(removeFromEnd: true, removeFromStart: false);
+                    break;
+                case "9":
+                    //Prompt to get index of user to remove from array
+                    PromptRemoveUser(removeFromEnd: false, removeFromStart: false);
+                    break;
+                case "10":
+                    //Sorting array by usernames in ascending order
+                    SortUsers();
+                    break;
+                case "11":
+                    LogOut();
+                    break;
+                default:
+                    MenuSelection();
+                    break;
+            }
+        }
+
+        static string PromptInput(string message) {
+            Console.WriteLine(message);
+            return Console.ReadLine();
+        }
+
+        //Login function - gets username/password and checks against array
+        static void Login() {
+            Console.Clear();
+            string username = PromptInput("Enter username:");
+            string password = PromptInput("Enter password:");
+
+            if (CheckIfUserExists(username))
+            {
+                Console.WriteLine("userexists");
+                string[] userDetails = ReturnUserByUsername(username);
+                Console.WriteLine(userDetails[0] + " " + userDetails[1]);
+                if (userDetails[1] == password) {
+                    loggedIn = true;
+                    
+                    MenuSelection();
+                    return;
+                }
+            }
+
+            if (RetryConfirmation("Bad username or password."))
+            {
+                Login();
+            }
+            else {
+                MenuSelection();
+            }
+        }
+
+        static void LogOut() {
+            Console.Clear();
+            loggedIn = false;
+            MenuSelection();
+        }
+
+        // Function to display error message and return true for retry;
+        static bool RetryConfirmation(string messageToDisplay) {
+            Console.Clear();
+            Console.WriteLine(messageToDisplay+" Retry? y/n");
+
+            if (Console.ReadLine().ToLower() == "y")
+            {
+                return true;
+            }
+            return false;
+        }
+
+
+        // Returns true if username is found
+        static bool CheckIfUserExists(string username) {
+            for (int i = 0; i < userArray.Length; i++)
+            {
+                if (userArray[i][0].ToLower() == username.ToLower())
+                {
+                    Console.WriteLine("Found");
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        // Look through user array for a username and return string array of username and password
+        static string[] ReturnUserByUsername(string username) {
+            string[] array = new string[] {"",""};
+            for (int i = 0; i < userArray.Length; i++)
+            {
+                if (userArray[i][0].ToLower() == username.ToLower())
+                {
+                    array = userArray[i];
+                }
+            }
+            return array;
+        }
+
+        //Adding string[] - [username,password] - user details to userArray at selected index.
+        static void AddUserToArrayAtIndex(string[] userDetails, int index) {
+            string[][] newArray = new string[userArray.Length + 1][];
+
+            for (int i = 0; i < newArray.Length; i++)
+            {
+                if (index == 0 && i == 0 )
+                {
+                    newArray[i] = userDetails;
+                    continue;
+                }
+
+                if (i < index)
+                {
+                    newArray[i] = userArray[i];
+                }
+
+                if (i==index)
+                {
+                    newArray[i] = userDetails;
+                }
+
+                if (i > index)
+                {
+                    newArray[i] = userArray[i - 1];
+                }
+            }
+
+            userArray = newArray;
+        }
+
+        //Function to list users;
+        static void ListUsers() {
+            Console.Clear();
+            Console.WriteLine("Users:\n");
+
+            for (int i = 0; i < userArray.Length; i++)
+            {
+                Console.WriteLine(userArray[i][0]);
+            }
+            PromptInput("\nPress any key to continue.");
+            MenuSelection();
+        }
+
+        //Function to add userDetails to end or start of userArray. Taking bool variables to determine whether add to start or end or select index.
+        static void PromptAddUserToArray(bool addToEnd = true,bool addToStart = false) {
+            Console.Clear();
+            Console.WriteLine("Adding user:\n");
+
+            string username = PromptInput("Enter username:");
+            string password = PromptInput("Enter password:");
+            
+            int index = userArray.Length;
+
+            if (addToStart)
+            {
+                index = 0;
+            }
+
+            if (!addToEnd && !addToStart)
+            {
+                string indexStr = PromptInput("Enter index to add user at:");
+                index = int.Parse(indexStr);
+            }
+
+            if (index<0 || index > userArray.Length)
+            {
+                if (RetryConfirmation("Bad index."))
+                {
+                    PromptAddUserToArray(addToEnd,addToStart);
+                }
+                else
+                {
+                    MenuSelection();
+                }
+            }
+
+            if (!CheckIfUserExists(username) || !(password == ""))
+            {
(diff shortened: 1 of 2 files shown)


Homework/2/Program.cs
```diff
@@ -0,0 +1,390 @@
...
+        
+
+        static void Main(string[] args)
+        {
+            //Adding a couple of users to array
+            userArray[0] = new string[] { "Admin", "slAptazodis" };
+            userArray[1] = new string[] { "user1", "pSwD" };
+            userArray[2] = new string[] { "LaIMONAS", "123" };
+            userArray[3] = new string[] { "Xavier", "Dust" };
+            userArray[4] = new string[] { "Blaber", "Lol" };
```
Almantask: Some cases don't add to anything extra and few of them would be enough to check if capitalization is truly being ignored.

Almantask: Be minimalistic.

Homework/2/Program.cs
```diff
@@ -0,0 +1,390 @@
...
+                            temp = userArray[i];
+                            userArray[i] = userArray[i + 1];
+                            userArray[i + 1] = temp;
+                        }
+                    }
+                   
+                }
+            }
+            ListUsers();
+        }
```
Almantask: Please note that this part wasn't really required, but I assume you took the homework assignment and wanted to apply some context to it, which is not a bad thing.

Homework/2/Program.cs
```diff
@@ -0,0 +1,390 @@
...
+            {
+                if (RetryConfirmation("Bad sort selected"))
+                {
+                    SortUsers();
+                }
+                else
+                {
+                    MenuSelection();
+                }
+            }
```
Almantask: split into 1 method

Homework/2/Program.cs
```diff
@@ -0,0 +1,390 @@
...
+                        if (userArray[i][0].CompareTo(userArray[i + 1][0]) < 0)
+                        {
+                            temp = userArray[i];
+                            userArray[i] = userArray[i + 1];
+                            userArray[i + 1] = temp;
+                        }
+                    }
+                   
+                }
+            }
```
Almantask: And then second

Homework/2/Program.cs
```diff
@@ -0,0 +1,390 @@
...
+            {
+                for (int i = 0; i < userArray.Length - 1; i++)
+                {
+                    if (order == "1")
+                    {
+                        if (userArray[i][0].CompareTo(userArray[i + 1][0]) > 0)
+                        {
+                            temp = userArray[i];
+                            userArray[i] = userArray[i + 1];
+                            userArray[i + 1] = temp;
```
Almantask: Could have a method `Swap`, duplicate code here and below.

Homework/2/Program.cs
```diff
@@ -0,0 +1,390 @@
...
+            {
+                if (RetryConfirmation("Bad index."))
+                {
+                    PromptRemoveUser(removeFromEnd, removeFromStart);
+                }
+                else
+                {
+                    MenuSelection();
+                }
+            }
```
Almantask: This block of code should be in a new function called `PromptIndex`. That way, you would have a high level function which: gets index, removes user at that index and refreshes the list on screen. This makes readability of this function a ton better, doesn't it? Right now, with the current situation, you mix the above code with high level calls like `RemoveUser` and `ListUsers` which does makes this function do more than one thing- which it shouldn't.

Homework/2/Program.cs
```diff
@@ -0,0 +1,390 @@
...
+            for (int i = 0; i < userArray.Length; i++)
+            {
+                Console.WriteLine(userArray[i][0]);
+            }
+            PromptInput("\nPress any key to continue.");
+            MenuSelection();
+        }
+
+        //Function to add userDetails to end or start of userArray. Taking bool variables to determine whether add to start or end or select index.
+        static void PromptAddUserToArray(bool addToEnd = true,bool addToStart = false) {
```
Almantask, line 298: Same remarks as in remove user.

Homework/2/Program.cs
```diff
@@ -0,0 +1,390 @@
...
+
+                if (i==index)
+                {
+                    newArray[i] = userDetails;
+                }
+
+                if (i > index)
+                {
+                    newArray[i] = userArray[i - 1];
+                }
```
Almantask: Only one of these conditions shuold be true at a time.
Therefore, use `else if` and not `if`. This will also allow you to remove the continue statement, because the if that is hit will stop checking others further ifs.

Homework/2/Program.cs
```diff
@@ -0,0 +1,390 @@
...
+                    //Remove from end of the array
+                    PromptRemoveUser(removeFromEnd: true, removeFromStart: false);
+                    break;
+                case "9":
+                    //Prompt to get index of user to remove from array
+                    PromptRemoveUser(removeFromEnd: false, removeFromStart: false);
+                    break;
+                case "10":
+                    //Sorting array by usernames in ascending order
+                    SortUsers();
```
Almantask: Do those comments add any value? They are awfully similar to the method names. Maybe you could clarify method names instead?

Almantask (changes requested):
Not bad overall, great effort, but please try to make amends based on the remarks to slightly improve it.

Homework/2/Program.cs
```diff
@@ -49,62 +64,112 @@ static void Main(string[] args)
...
+                        if (CheckIfUserExists(userCredentials[0]))
+                        {
+                            throw new DuplicateUserCredentialsException(userCredentials[0]);
+                        }
+                        else
+                        {
+                            AddUserToArrayAtIndex(userCredentials, 0);
+                        }
+                    }
+                }
```
Almantask: Could move this into a function of its own, will be more clear what are you actually trying.

Almantask (changes requested):
Is there `UsersNotFoundException`?

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+        //If a Users.text file contains duplicate users, throw a DuplicateUserCredentialsException.
+        //
+        //Restrictions
+        //Don't use File class static methods.
+
+
+
+        //Variables
+        private static bool loggedIn = false; // Variable to track if user is "logged in"
+        private static string[][] userArray = new string[0][]; // array of user arrays with login info which is string[username,password]; Would rather use List<>
```
westonpace: If you are going to say `Would rather use List<>` you should explain why you cannot use `List<>`.  Here it's just a homework assignment and I'm guessing it's because you are using `userArray` for your HW2 work but in the future it might not be so easy to understand why.

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+        private static string pathToFile = @"./users.txt"; // path to user file
```
westonpace: Are you on Windows?  You should use `.\users.txt` if so.  If not, then this is fine.  Even better, you can just say `users.txt` and your code will work on any OS.

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+                    while (!reader.EndOfStream)
+                    {
+                        string[] userCredentials = reader.ReadLine().Split(';');
+                        if (CheckIfUserExists(userCredentials[0]))
+                        {
+                            throw new DuplicateUserCredentialsException(userCredentials[0]);
+                        }
+                        else
+                        {
+                            AddUserToArrayAtIndex(userCredentials, 0);
```
westonpace: This will put the users into `userCredentials` in reverse order.  It's not technically a bad thing but it seems like an unneeded complication.  Best to add each new user to the end of the array.

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+                        }
+                    }
+                }
+            }
+            catch (FileNotFoundException fileEx)
+            {
```
westonpace, line 86: Never have an empty catch block.  Not even in a practice program or a quick program you throw together.  Don't catch the exception at all or, at least, print `fileEx`.

You can break this rule only if you really don't care if you get an exception or not.  In that case you should add a comment explaining this.  For example, in some languages (not C#) the "MakeDirectory" method throws an exception if the directory already exists...

```
try
{
  MakeDirectory("WorkDirectory");
}
catch (DirectoryAlreadyExistsException)
{
  // This is ok, we can continue if the directory already exists
}
```

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+            }
+            catch (DuplicateUserCredentialsException usersEx)
```
westonpace, line 89: Don't catch an exception just to rethrow it like this.  Just don't catch it in the first place and you achieve the same effect with less code.

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+        //Login function - gets username/password and checks against array
+        static void LogIn() {
+            Console.Clear();
+            string username = PromptInput("Enter username:");
+            string password = PromptInput("Enter password:");
+
+            if (CheckIfUserExists(username))
+            {
+                string[] userDetails = ReturnUserByUsername(username);
+                Console.WriteLine(userDetails[0] + " " + userDetails[1]);
```
westonpace: Hopefully this line was for debugging purposes :).  Don't want to dump `userDetails[1]` to the console before you've even checked their password.

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+                if (userDetails[1] == password) {
+                    loggedIn = true;
+                    MenuSelection();
+                    return;
+                }
+            }
+
+            if (RetryConfirmation("Bad username or password."))
+            {
+                LogIn();
```
westonpace, line 216: Technically recursing without bound in this way could lead to something called a "stack overflow" but it seems pretty unlikely in this case.  So just take this as an interesting note.

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+            if (Console.ReadLine().ToLower() == "y")
+            {
+                return true;
+            }
+            return false;
+        }
+
+
+        // Returns true if username is found
+        static bool CheckIfUserExists(string username) {
```
westonpace: This method and `ReturnUserByUsername` are very similar.  If you changed `ReturnUserByUsername` to return `null` if it didn't find anything then this method could become...

```
static bool CheckIfUserExists(string username) {
  return ReturnUserByUsername(username) != null;
}
```

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+            return array;
+        }
+
+        //Adding string[] - [username,password] - user details to userArray at selected index.
+        static void AddUserToArrayAtIndex(string[] userDetails, int index) {
+            string[][] newArray = new string[userArray.Length + 1][];
+
+            for (int i = 0; i < newArray.Length; i++)
+            {
+                if (index == 0 && i == 0 )
```
westonpace, line 266: This `if` block, while harmless, is not needed.

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+        }
+
+        //Function to list users;
+        static void ListUsers() {
+            Console.Clear();
+            Console.WriteLine("Users:\n");
+
+            for (int i = 0; i < userArray.Length; i++)
+            {
+                Console.WriteLine(userArray[i][0]);
```
westonpace, line 291: Here's a spot where it might be nice to have a leading space or tab or " - ".  Something to help read the list.

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+                }
+            }
+
+            userArray = newArray;
+        }
+
+        //Function to list users;
+        static void ListUsers() {
+            Console.Clear();
+            Console.WriteLine("Users:\n");
```
westonpace, line 287: In general, try to avoid `\n` in favor of `Environment.NewLine` but it is not as big of a deal these days.

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+
+            if (!CheckIfUserExists(username) && !(password == "") && index != -1)
+            {
+                string[] userDetails = new string[2] { username, password };
+                AddUserToArrayAtIndex(userDetails, index);
+                WriteUsersToFile();
+                ListUsers();
+            }
+            else {
+                if (RetryConfirmation("Bad user details."))
```
westonpace, line 316: Well, it might be a bad position and good user details.  Also, what does "bad user details" mean?

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+                    MenuSelection();
+                }
+            }
+        }
+
+        //Function to get index of where user should be removed
+        static void PromptRemoveUser(bool removeFromEnd = true, bool removeFromStart = false) {
+            Console.Clear();
+            int index = GetIndex(removeFromEnd, removeFromStart);
+            if (index!=-1)
```
westonpace: Probably nice to do something in an else here like tell the user they entered a bad index.

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+                    PromptRemoveUser(removeFromEnd: false, removeFromStart: false);
+                    break;
+                case "10":
+                    SortUsers();
+                    break;
+                case "11":
+                    LogOut();
+                    break;
+                case "12":
+                    WriteUsersToFile();
```
westonpace, line 184: You call `WriteUsersToFile` every time you add a user to a file and everytime you remove a user from a file so why do you need to call it here?

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+        //Remove user from userArray at index.
+        static void RemoveUser(int index) {
+            string[][] newArray = new string[userArray.Length - 1][];
+
+            for (int i = 0; i < newArray.Length; i++)
+            {
+                if (i <index)
+                {
+                    newArray[i] = userArray[i];
+                }else if (i == index)
```
westonpace: This block and the next block are identical.  Seems you could simplify.

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+            temp = userArray[x];
+            userArray[x] = userArray[y];
+            userArray[y] = temp;
+        }
+
+        //Ascending order sorting function
+        static void SortUserArrayInAscOrder() {
+            for (int j = 0; j < userArray.Length; j++)
+            {
+                for (int i = 0; i < userArray.Length - 1; i++)
```
westonpace, line 414: This is a bubble sort but it could be optimized.  Generally when doing a bubble sort your inner loop takes advantage of the fact that each iteration can be shorter (because you've put an item correctly at the end).

Another potential optimization is to bail out early if you didn't swap any items.

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+        // Function to swap values in userArray
+        static void UserArrayValueSwap(int x, int y) {
+            string[] temp;
+            temp = userArray[x];
+            userArray[x] = userArray[y];
+            userArray[y] = temp;
+        }
+
+        //Ascending order sorting function
+        static void SortUserArrayInAscOrder() {
```
westonpace: This is almost identical to `SortUserArrayInDescOrder`.   You could pass a boolean arg in instead.

Homework/2/Program.cs
```diff
@@ -0,0 +1,470 @@
...
+                    if (userArray[i][0].CompareTo(userArray[i + 1][0]) < 0)
+                    {
+                        UserArrayValueSwap(i, i + 1);
+                    }
+                }
+            }
+        }
+
+        //True for ascending/ false for descending
+        static bool ReturnSortingOrder()
```
westonpace, line 442: We haven't learned about it yet but it might be nice to return an enum here to keep the code more readable.  For example, is "false" descending or ascending?  Another way to tackle it might be to rename this method to something like `GetShouldSortAscending`

westonpace (commented):
It's good.  I do appreciate that you prompt with numbers in your menus so it's clear what the choices are.   In some places it seems like you rely on static variables instead of passing arguments.

Also, I think I mentioned somewhere but the general flow of this is very recursive and could lead to stack overflow and also make debugging complicated (since stack traces might grow rather large) but that isn't such a big deal when you're starting out.

Homework/2/Program.cs
Laimonukas: I thought since UsersNotFoundExecption was called if no users.txt file was found, that it would be ok if I used FileNotFoundExecption instead.

Added UsersNotFoundException exeption an moved reader to it's own function.

Homework/2/Program.cs
Laimonukas: I wanted to have something like autosave so I wouldn't have to add users every time in case I messed up somewhere.

Homework/2/Program.cs
Laimonukas: It's just some default error message, I didn't want to deal with all false cases.

Homework/2/Program.cs
Laimonukas: Added else if block with retry.

Homework/2/Program.cs
Laimonukas: Removed unneeded else if.

Homework/2/Program.cs
Laimonukas: Combined functions to accept bool argument

Homework/2/Program.cs
Laimonukas: Tried optimizing this sort.

Almantask:
expired

PR #41, opened 2021-04-05
Homework/3

cjahermanns:
One thing I noticed was there should be some kind of mechanism to only append to ```Users.txt``` when the writer is at the beginning of a new line.


Joedmin:
> 
> 
> One thing I noticed was there should be some kind of mechanism to only append to `Users.txt` when the writer is at the beginning of a new line.

Just make sure the users.txt will be ending with a new line when you open it with a stream, then it's all as it should be

cjahermanns:
> Just make sure the users.txt will be ending with a new line when you open it with a stream, then it's all as it should be

I tried within the append method but I couldn't get it to work... I got an error when I tried to ```fs.Read()``` the length of an ```byte[]``` containing ```Environment.NewLine``` and comparing the result with the array.

Will give it another try during the week though :D

HW3/Exceptions/UsersNotFoundException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.IO;
+
+namespace HW3.Exceptions
+{
+    internal class UsersNotFoundException : FileNotFoundException
+    {
+        public UsersNotFoundException(string? message, string? filename) : base(message, filename)
```
Almantask: Maybe a full message is not needed, when you can form it out of a missing filename? You could at least make it optional and there doesn't seem to be much value in always providing a different one.

HW3/Exceptions/DuplicateUserCredentialsException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace HW3.Exceptions
+{
+    internal class DuplicateUserCredentialsException : Exception
+    {
+        public DuplicateUserCredentialsException(string? message) : base(message)
```
Almantask: Do you know the difference betwen `string` and `string?`

HW3/IO/Write.cs
```diff
@@ -0,0 +1,39 @@
...
+                using (var sw = new StreamWriter(fs))
+                {
+                    sw.WriteLine($"{username}\t{password}");
+                }
+            }
+        }
+
+        private static void CheckForNewLine(string filename)
+        {
+            string contents;
```
Almantask: declare the variable closer to its usage.

HW3/IO/Write.cs
```diff
@@ -0,0 +1,39 @@
...
+            using (var fs = new FileStream(filename, FileMode.Append))
+            {
+                using (var sw = new StreamWriter(fs))
+                {
+                    sw.WriteLine($"{username}\t{password}");
+                }
+            }
+        }
+
+        private static void CheckForNewLine(string filename)
```
Almantask: Could you please explain the purpose of this method. I don't understand why do you have both reader and writer in the same method.

HW3/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+{
+    internal class Program
+    {
+        private const string usersFile = "Users.txt";
+
+        private static bool ConsoleLogin()
+        {
+            try
+            {
+                string username = ConsoleReadLineWithTextPrompt("Username:\t");
```
Almantask: A method name does not have to have all those details. Simply call it: `PromptText`

HW3/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+                    {
+                        throw new UsersNotFoundException(e.Message, usersFile);
+                    }
+
+                    return;
+                }
+
+                Console.WriteLine("Username already exists! Please try again.");
+            } while (true);
+        }
```
Almantask: Does too much, pleas esplit.

HW3/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+
+                    case ConsoleKey.E:
+                        return;
+
+                    default:
+                        Console.WriteLine($"{keyPressed} is not an option. Try again!");
+                        break;
+                };
+            } while (!loggedIn);
+        }
```
Almantask: Don't have this much logic in main. Split into lower level functions which main simply calls.

Almantask (changes requested):
Also change target branch to main.

HW3/Exceptions/DuplicateUserCredentialsException.cs
westonpace: Every time you create this exception you pass the value `"Error: Duplicate userentry detected!"`.  It might be simpler to just do...

```
internal class DuplicateUserCredentialsException : Exception
{
  public DuplicateUserCredentialsException() : base("Error: Duplicate userentry detected!")
  {
  }
}
```

HW3/Exceptions/DuplicateUserCredentialsException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace HW3.Exceptions
+{
+    internal class DuplicateUserCredentialsException : Exception
```
westonpace: You should probably avoid marking exceptions `internal` unless you have a good reason.

HW3/IO/Read.cs
```diff
@@ -0,0 +1,49 @@
...
+                            }
+
+                            users.Add(splittedline[0], splittedline[1]);
+                        }
+                    }
+                }
+            }
+            catch (FileNotFoundException ex)
+            {
+                throw new UsersNotFoundException(ex.Message, filename);
```
westonpace: Hmm...you are catching `FileNotFoundException` and then wrapping it in `UsersNotFoundException` which extends `FileNotFoundException`.  It all feels a little janky to me.  Also, `ex` already has the filename in it if it is thrown by system methods.

If you ever want to catch an exception and either add more information to it (e.g. "This file should have come as part of the installation and should not be modified, please reinstall") or if you want to wrap a system exception with your own particular invocation (some APIs do this to reduce the number of potential exceptions that could be thrown by a library) then it is better to do what is called an "inner exception".  Something like...
```
class UsersNotFoundException : Exception
{
  public UsersNotFoundException(FileNotFoundException inner) : base("Could not find the users database file", inner)
  {
  }
}
```

HW3/IO/Read.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace HW3.IO
+{
+    public static class Read
+    {
+        public static Dictionary<string, string> UserCredentials(string filename)
+        {
+            Dictionary<string, string> users = new Dictionary<string, string>();
```
westonpace, line 11: Good thought to use a `Dictionary` here.

HW3/IO/Read.cs
```diff
@@ -0,0 +1,49 @@
...
+                        {
+                            string line = sr.ReadLine();
+
+                            if (string.IsNullOrWhiteSpace(line)) continue;
+
+                            string[] splittedline = line.Split("\t");
+
+                            if (users.ContainsKey(splittedline[0]))
+                            {
+                                throw new DuplicateUserCredentialsException("Error: Duplicate userentry detected!");
```
westonpace: It's a little late if we only detect this when reading the file.  Best to detect this before we write out the bad file.  Now the user has to go and repair their users file.

HW3/IO/Read.cs
```diff
@@ -0,0 +1,49 @@
...
+                {
+                    using (var sr = new StreamReader(fs))
+                    {
+                        while (!sr.EndOfStream)
+                        {
+                            string line = sr.ReadLine();
+
+                            if (string.IsNullOrWhiteSpace(line)) continue;
+
+                            string[] splittedline = line.Split("\t");
```
westonpace: I'm not the biggest fan of the name `splittedline` but if you are going to use it then use `splittedLine`.  Personal opinion: when I do splits like this I tend to name the variable something like `usernamePassword`.

HW3/IO/Write.cs
```diff
@@ -0,0 +1,39 @@
...
+            using (var fs = new FileStream(filename, FileMode.Open))
+            {
+                var sr = new StreamReader(fs);
+                contents = sr.ReadToEnd();
+
+                if (!contents.EndsWith(Environment.NewLine))
+                {
+                    var sw = new StreamWriter(fs);
+                    sw.Write(Environment.NewLine);
+                    sw.Flush();
```
westonpace: Flush is automatically called when a `StreamWriter` is disposed.  Speaking of dispose...I don't see a using here.  You can "get away" with it when using `FileStream` (since it's really only important that at least one of them be disposed).  Also, if you disposed the `StreamReader` this method would actually fail....hmmm....

I think I would open a `FileStream` twice.  First, open it and read it like you did (except dispose the `StreamReader`).  Then open it in append mode if the newline is needed.  But I think your approach is ok too.

HW3/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                string username = ConsoleReadLineWithTextPrompt("Username:\t");
+
+                string password = ConsoleReadLineWithTextPrompt("Password:\t");
+
+                return VerifyCredentials(username, password);
+            }
+            catch (UsersNotFoundException ex)
+            {
+                Console.WriteLine($"Couldn't load user credentials. Please contact @Kaisinel: {ex.Message}");
```
westonpace: Poor kai, getting so much spam 😆 

Also, I'm not sure I'm a fan of this catch statement.  Consider the logic.  First you catch and print this message.  Then you return false so you have to check that return and you print an even more vague message (oh no something went wrong).  Then you loop and you start the program all over again but there is no way to recover from this so things are just going to go wrong again.

On the other hand, you could make "Couldn't load user credentials. Please contact @Kaisinel" the default error message for `UsersNotFoundException` and then don't catch it anywhere.  The program will crash (as it should) and you don't have to return a boolean here so there is no "oh no something went wrong" and the program is simpler.

HW3/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+                return false;
+            }
+        }
+
+        private static string ConsoleReadLineWithTextPrompt(string promptText)
+        {
+            Console.Write(promptText);
+            string username = Console.ReadLine();
```
westonpace: Don't create a variable just to return it.  Prefer `return Console.ReadLine();`

HW3/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+            return username;
+        }
+
+        private static void ConsoleRegisterAccount()
+        {
+            Dictionary<string, string> users;
+            try
+            {
+                users = IO.Read.UserCredentials(usersFile);
```
westonpace: Nice use of multiple classes and namespaces.

HW3/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+            catch (UsersNotFoundException ex)
```
westonpace: Same comments as the last time you caught this exception.

HW3/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                Console.WriteLine($"Couldn't load user credentials. Please contact @Kaisinel: {ex.Message}");
+                return;
+            }
+
+            do
+            {
+                string username = ConsoleReadLineWithTextPrompt("Username:\t");
```
westonpace: Nit: In some places you have newlines where I would not.  Try to avoid making your program look like you double spaced it.  Spaces in code are valuable markers if you use them judiciously.  If you use too many then you lose the effect.

HW3/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+                    {
+                        IO.Write.AppendCredentials(usersFile, username, password);
+                        Console.WriteLine("Useraccount created! You may login now.");
+                    }
+                    catch (FileNotFoundException e)
+                    {
+                        throw new UsersNotFoundException(e.Message, usersFile);
+                    }
+
+                    return;
```
westonpace: Put this `return` inside the `try` block for better readability.

HW3/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+
+                Console.WriteLine("Username already exists! Please try again.");
+            } while (true);
+        }
+
+        private static void Main(string[] args)
+        {
+            bool loggedIn = false;
+
+            Console.WriteLine("Hello!");
```
westonpace: Hi.

HW3/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            do
+            {
+                Console.WriteLine("Please select an option:");
+                Console.WriteLine("[L]ogin - [R]egister - [E]xit");
+
+                var keyPressed = Console.ReadKey(true).Key;
```
westonpace, line 54: Nit: The user is not entering a password here.  Let them see what they entered.

HW3/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+                        break;
+                };
+            } while (!loggedIn);
+        }
+
+        private static bool VerifyCredentials(string username, string password)
+        {
+
+                var users = IO.Read.UserCredentials(usersFile);
+                return users.ContainsKey(username) && users[username].Equals(password);
```
westonpace: Technically this is going to require a case sensitive username match.  There is a way you can use a dictionary and have case insensitive match.  Can you think what it might be?

westonpace (commented):
Good submission.  I think the exception handling could be cleaned up a bit (less is more) but not bad.

HW3/Exceptions/DuplicateUserCredentialsException.cs
cjahermanns: Good point! Yes, I know the difference between string and string?... string? is nullable.

HW3/IO/Read.cs
cjahermanns: That's a nice way I guess... Will try to embrace this naming convention :D

HW3/IO/Write.cs
cjahermanns: I figured, if the Users.txt is not written by the application but instead comes from another source, I shouldn't assume it ending on a new line and thus appending a new username/password would extend the last password in the list. So that was my rushed approach to make sure I append new credentials in one line each.

HW3/IO/Read.cs
cjahermanns: Ty ^^

HW3/IO/Read.cs
cjahermanns: That's right, but what if the Users.txt was not created by the application or it got manipulated?

HW3/Program.cs
cjahermanns: > The program will crash (as it should)

Well actually the assignment was ```If there is no Users.txt file- the application should not throw `UsersNotFoundException`.```. I thought there should be a specific exception but it should not been thrown. I wanted to inform the user instead of crashing the app.

HW3/Program.cs
cjahermanns: I know I'm not consistent with new lines... I try to seperate everything with 1 line (except for fields wich are nicely grouped together). I use CodeMaid to get rid of any extra lines but sometimes I clutter everything together and simply forget that CM only removes extra lines but does not insert them. Maybe I should reconsider this habit...

HW3/Program.cs
cjahermanns: I suppress the userinput because the console stays on that line when I `WriteLine()` what to do next...

HW3/Program.cs
cjahermanns: I guess I could loop through the key value pairs and compare `Key.ToLower()` with `username.ToLower()` but that wouldn't be very efficient and I wouldn't need the dictionary. Maybe I could use a class deriving from `Dictionary` with an own `IEqualityComparer` implementation.

cjahermanns:
> Also change target branch to main.

Did I do this correctly? I edited the PR to target main.

Almantask:
> > Also change target branch to main.
> 
> Did I do this correctly? I edited the PR to target main.

Yes, you did it correct.

Almantask (approved):

PR #43, opened 2021-04-06
HW3: implement a credentials management system

gregory-butler:
LE:
I refactored the solution using lists and OOP.
I implemented the following classes:

Credential:
 - it holds the the user name and password.
 - it implements IEquatable 
 - implemented TryParse pattern
 - and ToString override method

MenuItem:
 - it holds the menu id, label, associated console key and action to invoke
 - ToString override method

Menu: 
- it the holds the menu header, a list of menu items and a menu item designated for the exit of the program.
- I implemented an indexer which returns an menu item when the correct key is found
- ToString overide method.

No TDD for this part, still learning about OOP.

--------------------------------------------------------------
Old comment before commit [cce7601](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/a9ed77d0f7138da573b3c30c80f235300aabbde6)

I implemented methods that try to replicate the functionality of those in the 'File' class which I store in a static class called 'Files'. For these I also implemented tests and I tried to do them in a TDD fashion (I tried to respect this cycle: Write a failing unit test, Make the test pass, Refactor)

For the Login Register parts of the homework I created a class called 'Credentials' in which I implemented the Login and Register methods as public and private helper methods. To get and store the credentials I made use of the methods from my previously implemented 'Files' class.

For the actual user interaction I created a new class called 'App' which exposes only one method called 'Run'. This class is more of an UI for the user of the application. Under the hood I created 3 arrays which hold the menu items, the keys to be pressed to access the menu item and the actions that will start when the corresponding menu item is selected.
To add a new menu item I would just have to implement a method for it and add the entries to the corresponding  arrays.


CredentialsManager/App.cs
```diff
@@ -0,0 +1,161 @@
...
+namespace CredentialsManager
+{
+  public static class App
+  {
+    private static readonly Menu MainMenu = CreateMenu();
+
+    private const string Header = "Credentials Manager";
+    private const string LoginMessage = "Welcome";
+    private const string FailedLoginMessage = "Invalid credentials!";
+    private const string RegistrationSuccessfulMessage = "Registration successful!";
```
Almantask: Cool use of constants.

CredentialsManagerTests/FilesTests.cs
```diff
@@ -0,0 +1,112 @@
...
+      Action actual = () => Files.WriteLine(nonexistentFile, string.Empty, append);
+
+      actual.Should().ThrowExactly<MyFileNotFoundException>(nonexistentFile);
+
+    }
+    
+    [Theory]
+    [InlineData(@"Files/Output/expected_single_line.txt", "AllTheSingleLines", false)]
+    [InlineData(@"Files/Output/expected_multi_line.txt", "Line1", true)]
+    public void WriteLine_Should_Create_Expected_Output(string file, string expected, bool append)
```
Almantask: You say you are testing WriteLine, but in the actual line is ReadAllLines. It makes it very confusing.
Try using 
```
// Arrange

// Act

// Assert
```
for when it's hard to understand what is actually being tester. My rule of thumb is to separe arrange, act and assert ea h with a blank line, but keep the rest in the same code block.

CredentialsManagerTests/FilesTests.cs
```diff
@@ -0,0 +1,112 @@
...
+      File.Delete(file);
+      Files.WriteAllLines(file, expected);
+
+      var actual = File.ReadAllLines(file);
+
+      actual.Should().Equal(expected);
+    }
+
+    [Theory]
+    [InlineData(true)]
```
Almantask: Append should not determine the outcome, therefore you can hardcode it (but hardcode it like `anyBool = true`;

CredentialsManagerTests/FilesTests.cs
```diff
@@ -0,0 +1,112 @@
...
+      var actual = File.ReadAllText(file);
+
+      actual.Should().Be(expected);
+    }
+
+    [Theory]
+    [InlineData(@"Files/Output/expected_lines1.txt", new[] { "Test.l", "Line2.l" })]
+    [InlineData(@"Files/Output/expected_lines_empty.txt", new string[] {})]
+    [InlineData(@"Files/Output/expected_lines_empty2.txt", new[] { "", "", "" })]
+    public void WriteAllLines_Should_Create_Expected_Output(string file, string[] expected)
```
Almantask: Your test name should specify what that expected output is.

CredentialsManagerTests/FilesTests.cs
```diff
@@ -0,0 +1,112 @@
...
+using CredentialsManager;
+using FluentAssertions;
+using System.IO;
+using System.Linq;
+using CredentialsManager.FilesExceptions;
+using Xunit;
+
+namespace CredentialsManagerTests
+{
+  public class FilesTests
```
Almantask: Test names are very redundant. Consider adding two nested classes and per each method under test- makes test name less redundant. For example:
```cs
public class FilesTests
{
    public class ReadaAllText
    {
        Should_Throw_MyFileNotFoundException_When_Path_Is_Not_Present_Or_Not_Able_To_Open()
        {
          //...
        }
    }
}
```

Almantask: Shorter names and grouped by method by design ;)

CredentialsManager/Users.txt
```diff
@@ -0,0 +1,5 @@
+﻿kai,1234
```
Almantask: That's me!

CredentialsManager/Program.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace CredentialsManager
+{
+  internal static class Program
+  {
+    private static void Main(string[] args)
+    {
+      App.Run();
```
Almantask, line 7: I like this :)

CredentialsManager/Models/MenuItem.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Threading.Tasks;
+
+namespace CredentialsManager.Models
+{
+  public class MenuItem
+  {
+    public int ItemId { get; set; }
+    public string Label { get; set; }
+    public ConsoleKey ConsoleKey { get; set; }
+    public Action Action { get; set; }
```
Almantask: Since they won't ever change upon initialization, consider passing all through ctor.

CredentialsManager/Models/Credentials.cs
```diff
@@ -0,0 +1,100 @@
...
+      if (string.IsNullOrWhiteSpace(userPassword))
+      {
+        throw new ArgumentException("Value can not be null or empty", nameof(userName));
+      }
+      
+      UserName = userName;
+      UserPassword = userPassword;
+    }
+
+    public static bool TryParse(string line, string delimiter, out Credentials credentials)
```
Almantask, line 20: Also neat

CredentialsManager/Models/Credentials.cs
```diff
@@ -0,0 +1,100 @@
+﻿using System;
+
+namespace CredentialsManager.Models
+{
+  public class Credentials : IEquatable<Credentials>, IComparable<Credentials>, IComparable
```
Almantask: Can be a struct instead of a class, because it is a `ValueObject` essentially, just 2 fields which you treat as 1.

Almantask (changes requested):
Very high effort homework, high quality code. Please try to make tests better by removing unnecessary details and organizing code as recommended.

CredentialsManager/Program.cs
gregory-butler: Thank you, I like it too :)

CredentialsManager/Users.txt
gregory-butler: yes! :D

westonpace:
Minor thing: "implement a credentials management system" is a great title for a real world pull request.  However, for this boot camp, maybe just call it "HW3" so it is more clear to use reviewers what you are submitting.

CredentialsManager/App.cs
westonpace: I was just going to comment "I'm not a huge fan of single-use string constants" 😆 .  To each their own.  In a big enterprisy application these would be replaced by lookups in some kind of localization database.

CredentialsManager/App.cs
```diff
@@ -0,0 +1,161 @@
...
+
+    private static bool _isExitInvoked;
```
westonpace: This and `Exit` seem unused?

CredentialsManager/App.cs
```diff
@@ -0,0 +1,161 @@
...
+        }
+        catch (MyFileNotFoundException ex)
+        {
+          Console.WriteLine(ex.Message);
+          break;
+        }
+        catch (DuplicateUserCredentialsException ex)
+        {
+          Console.WriteLine(ex.Message);
+          Pause("Press ENTER to exit..");
```
westonpace, line 38: Consistency.  Pause for all exceptions or none.  You don't pause on `MyFileNotFoundException`.

CredentialsManager/App.cs
```diff
@@ -0,0 +1,161 @@
...
+          break;
+        }
+        catch (UserNameIsTakenException ex)
```
westonpace: Don't use exceptions to replace normal control flow.

CredentialsManager/App.cs
```diff
@@ -0,0 +1,161 @@
...
+        {
+          Console.WriteLine(ex.Message);
+          Pause();
+        }
+      }
+    }
+
+    private static IList<MenuItem> GetMenuItems()
```
westonpace, line 44: 😮 You're getting sophisticated now.

CredentialsManager/App.cs
```diff
@@ -0,0 +1,161 @@
...
+    {
+      return new List<MenuItem>
+      {
+        new MenuItem { ItemId = 1, Label = "Login", Action = Login, ConsoleKey = ConsoleKey.D1 },
```
westonpace: This is very clever but one nit would be that `ConsoleKey.D1` isn't going to match `1` pressed on the numpad.  You should probably store a `char` here and then use `KeyChar` instead of `KeyCode` later.  Although that will be case sensitive😕 (not that you're using numbers here, just thinking general purpose) so I'd probably store the lowercase version here and then lowercase the input `KeyChar` when checking.

Also, `ItemId` seems redundant when you can just print `ConsoleKey`.

CredentialsManager/App.cs
```diff
@@ -0,0 +1,161 @@
...
+
+      Console.WriteLine(MainMenu.ToString());
+    }
+
+    private static void GetUserInput()
+    {
+      ConsoleKey pressedKey;
+      do
+      {
+        pressedKey = Console.ReadKey(true).Key;
```
westonpace, line 81: I'm a little torn about hiding the user's input here.  They aren't entering anything sensitive.

CredentialsManager/App.cs
```diff
@@ -0,0 +1,161 @@
...
+        if (!MainMenu.IsValidKey(pressedKey)) continue;
```
westonpace: This will just ignore invalid keys.  I think it will work here but my gut reaction would be to print "That was not a valid option" or something like that.

CredentialsManager/App.cs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    private static void PrintMenu()
+    {
+      ConsoleInit(Header);
+
+      Console.WriteLine(MainMenu.ToString());
+    }
+
+    private static void GetUserInput()
```
westonpace: Nit: You aren't just getting the user's input here.  You are also running the command.

CredentialsManager/App.cs
```diff
@@ -0,0 +1,161 @@
...
+      _isExitInvoked = true;
+    }
+
+    private static string PromptString(string message)
+    {
+      string userInput;
+      do
+      {
+        Console.Write(message);
+        userInput = Console.ReadLine() ?? string.Empty;
```
westonpace: `Console.ReadLine` will never return null during normal operation.  The only time it would return null is if the standard input stream is EOF.  In Windows this doesn't really happen unless you are launching a process as a subprocess of another parent process (in that case the input of the child process could be controlled directly by the parent process).

In Linux this happens when we redirect with the pipe `|` operator (technically you can also redirect on Windows, especially if you're a powershell junky).

Anyways, I bring this all up because...if `Console.ReadLine` returns `null` then `Console.ReadLine` is ALWAYS going to return `null` from here on out.  Consider what that would mean for your program here.

CredentialsManager/App.cs
```diff
@@ -0,0 +1,161 @@
...
+      } while (string.IsNullOrWhiteSpace(userInput));
+
+      return userInput;
+    }
+
+    private static string PromptString(string message, bool isKeyVisible)
```
westonpace: I like the cool tricks to hide the password.  I do not like the fact that you have two methods named `PromptString` like this.  It seems `PromptString(message)` is really just `PromptString(message, true)`.  Also, I don't think you call `PromptString(message, true)` anywhere in your code.

CredentialsManager/App.cs
westonpace: Ah, I see.  I was searching for `Exit()` but you are calling it as a menu action.  Your menu already has the concept of an `ExitItem`.  It seems that could be combined with a little logic to remove the need for `_isExitInvoked`.

CredentialsManager/Files.cs
```diff
@@ -0,0 +1,93 @@
...
+          return reader.ReadToEnd();
+        }
+      }
+      catch (Exception ex) when (ex is ArgumentException
+                              || ex is ArgumentNullException
+                              || ex is FileNotFoundException
+                              || ex is DirectoryNotFoundException
+                              || ex is IOException)
+      {
+        throw new MyFileNotFoundException($"{file} file not found ore not able to open!");
```
westonpace: You're kind of swallowing useful information here.  This whole catch / rethrow block feels a little iffy to me.  You could maybe just catch `FileNotFoundException` and translate it to provide a more helpful error message (although the `Message` set by `FileNotFoundException` is already pretty clear) but I wouldn't catch all of these.

Also, if you are going to catch and rethrow like this it is best to keep `ex` around as an "inner exception" (you can google this).  It keeps the stack trace of `ex` around instead of swallowing it.

CredentialsManager/Models/Credentials.cs
gregory-butler: I successfully converted to it to struct in commit [4b30e14](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/4b30e147718194bba663095be2c2d4fa23a90759)

CredentialsManager/Files.cs
```diff
@@ -0,0 +1,93 @@
...
+                              || ex is DirectoryNotFoundException
+                              || ex is PathTooLongException
+                              || ex is IOException
+                              || ex is SecurityException) 
+      {
+        throw new MyFileNotFoundException($"{file} file not found ore not able to open!");
+      }
+    }
+
+    public static void WriteAllLines(string file, string[] data)
```
westonpace, line 58: You never use this in your application.

CredentialsManager/FilesExceptions/DuplicateUserCredentialsException.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+
+namespace CredentialsManager.FilesExceptions
```
westonpace, line 3: It's a preference thing but this might be a bit too fine-grained of namespacing for my tastes.

gregory-butler:
> Minor thing: "implement a credentials management system" is a great title for a real world pull request. However, for this boot camp, maybe just call it "HW3" so it is more clear to use reviewers what you are submitting.

Thank you for the suggestion, I think I will add both, homework number in the form of a prefix like HW{hw_number) followed by the actual title. Sorry if I created confusion.

CredentialsManager/FilesExceptions/DuplicateUserCredentialsException.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+#pragma warning disable RCS1194 // Implement exception constructors.
+
+  public class DuplicateUserCredentialsException : Exception
+#pragma warning restore RCS1194 // Implement exception constructors.
+  {
+    public DuplicateUserCredentialsException()
```
westonpace: Get rid of this constructor (also, if you do, then you can probably get rid of that warning suppression).

CredentialsManager/Models/Credentials.cs
gregory-butler: thank you.

CredentialsManager/FilesExceptions/DuplicateUserCredentialsException.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+    }
+
+    public DuplicateUserCredentialsException(string? details) : base(details)
```
westonpace: Why `string?` and not `string`?

CredentialsManager/Models/MenuItem.cs
gregory-butler: I implement the constructor in commit [740594c](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/740594c30744a2df6e43ca5cc20e72d847cc0ef9)

CredentialsManager/FilesExceptions/MyFileNotFoundException.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace CredentialsManager.FilesExceptions
+{
+  public class MyFileNotFoundException : Exception
+  {
+    public MyFileNotFoundException()
+    {
+    }
+
+    public MyFileNotFoundException(string details) : base(details)
```
westonpace: Prefer...
```
public MyFileNotFoundException(string file) : base($"{file} file not found ore not able to open!")
{

}
```

westonpace: Also, you can make a static method named `FilenameToExceptionMessage` if you don't like putting that long string in the base constructor call.

CredentialsManager/Manager.cs
```diff
@@ -0,0 +1,86 @@
...
+  public static class Manager
+  {
+    private const string FieldDelimiter = ",";
+    private const string CredentialsFile = "Users.txt";
+    private static List<Credentials> _credentials = new();
+
+    public static void Initialize()
+    {
+      _credentials = GetCredentials();
+      _credentials.Sort();
```
westonpace: I was going to ask "why sort" but now I see that it is necessary for your efficient `CheckForDuplicates`.  In my own personal code this might be a place where I'd put an explanatory comment.  Something like `// Sorting here so CheckForDuplicates works properly`.  Or you could put the sort inside `CheckForDuplicates`.  However, it is not obvious why it is here.

CredentialsManager/Manager.cs
```diff
@@ -0,0 +1,86 @@
...
+      }
+
+      Files.WriteLine(CredentialsFile, CreateString(credentials), true);
+    }
+
+    private static bool IsUserNameTaken(Credentials credentials)
+    {
+      foreach (var storedCredential in _credentials)
+      {
+        if (storedCredential.UserName.Equals(credentials.UserName))
```
westonpace, line 73: You might want to do a case insensitive comparison here.

CredentialsManager/Models/Credentials.cs
westonpace: Or even a `record` these days since it is immutable.

CredentialsManager/Models/Credentials.cs
```diff
@@ -0,0 +1,100 @@
...
+      }
+
+      credentials = new Credentials(userName, userPassword);
+
+      return true;
+    }
+
+    public override string ToString()
+    {
+      return $"{UserName},{UserPassword}";
```
westonpace: This is helpful for debugging and this is just a toy homework application so no big deal.  However, in a real world application, you always have to be aware of how your password is being treated.  For example, this might accidentally lead to a password getting leaked by some debugger or exception or accidental print.

CredentialsManager/Models/Credentials.cs
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    public bool Equals(Credentials? other)
```
westonpace: I don't think the `?` is necessary here?  But I might be wrong.  I'm not up to speed on this operator.  Maybe it's needed for one of the interfaces.

CredentialsManager/Models/Credentials.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+      if (obj is null) return false;
+      if (ReferenceEquals(this, obj)) return true;
+
+      return obj.GetType() == GetType() && Equals((Credentials)obj);
+    }
+
+    public override int GetHashCode()
+    {
+      return HashCode.Combine(UserName.ToLowerInvariant(), UserPassword);
```
westonpace: If your `Equals` uses `StringComparison.Ordinal` and your `HashCode` uses `ToLowerInvariant` you can get subtle bugs with unicode string expansion since ordinal is not the same thing as invariant.

CredentialsManager/Models/Menu.cs
```diff
@@ -0,0 +1,73 @@
...
+
+    public MenuItem ExitItem { get; set; }
+
+    public Menu()
+    {
+    }
+
+    public Menu(string header, IList<MenuItem> menuItems, MenuItem exitItem)
+    {
+      Header = header ?? throw new ArgumentNullException(nameof(header));
```
westonpace: Personally I only do this kind of validation on input that comes from external sources (users, files, etc.)  Even if you later add the ability to (for example) create a menu from a file, you should be doing the validation when you read the menu file and not here.

CredentialsManager/Models/MenuItem.cs
westonpace: For C#9 you can do `{ get; init; }`.  Before C#9 you should at least do `{ get; private set; }`.  Or use `record`

westonpace (commented):
I didn't review the tests in the interest of time (and it looked like Kai already had a few things to say there) but this is a very well written submission.

CredentialsManager/App.cs
gregory-butler: I updated it in commit [a7686d9](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/a7686d9b086bc78617d5f8d28da7b7033c10bbff)

CredentialsManager/Manager.cs
gregory-butler: I updated this in commit [c269d01](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/c269d01381883751f5cdbbfb4ac2b879f6034f76)

CredentialsManager/App.cs
gregory-butler: I updated it in commit [abfdbdb](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/abfdbdb1584ff1e5f10ed1abbb2a0b43ac3d29ee)

CredentialsManager/App.cs
gregory-butler: could 'ExecuteUserInput' be a better name?

CredentialsManager/App.cs
gregory-butler: Thank you for the detailed explanation. This one was not intentional, I think I accepted a suggestion from the IDE (Rider) by mistake because after I deleted the coalesce operator it's suggesting to add it like you saw it here.

gregory-butler: I updated it in commit [57ba93c](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/57ba93c3d8d6eefbd10ab94be489528fef7569bf)

CredentialsManager/App.cs
gregory-butler: Is this a good thing?  My sarcasm detector does not work while online😃
I read that is best to use interfaces instead of the actual class.

CredentialsManagerTests/FilesTests.cs
gregory-butler: This is news to me, I actually did not know that you can nest classes 😲

CredentialsManager/App.cs
gregory-butler: I decided to remove the PromptString overload.
I updated in commit [17deff9](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/17deff93537954ed330d11ab0d6b08e1324dd225)

CredentialsManager/App.cs
gregory-butler: My intent was to not have an echo of the pressed key and to just execute the desired action of the menu item.

CredentialsManager/App.cs
gregory-butler: Inspecting the method revealed that Console.ReadLine() returns `string?`, this is why I was getting the suggestion.

CredentialsManager/FilesExceptions/DuplicateUserCredentialsException.cs
gregory-butler: I might not understand this correctly so I'm asking you:
Is the question why I used the nullable operator or you are asking me if I know to use it?
The reason it is there it's because I auto implemented the constructor in the IDE and this is the Exception class under the hood.
 ```
  public Exception(string? message, Exception? innerException)
            : this()
        {
            _message = message;
            _innerException = innerException;
        }
```
Microsoft is using nullabe operator everywhere

File: CredentialsManager/App.cs (hunk 17 above)
Comment (gregory-butler) [reply CredentialsManager/App.cs]: I personally don't like the idea but I made the update in commit [e8fd4d2](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/e8fd4d208d9e052397fc9c6bdc22c594b214b711)

File: CredentialsManager/App.cs (hunk 18 above)
Comment (gregory-butler) [reply CredentialsManager/App.cs]: Decided to go with 'ExecuteUserRequest'.
updated it in commit [e8fd4d2](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/e8fd4d208d9e052397fc9c6bdc22c594b214b711)

File: CredentialsManager/FilesExceptions/DuplicateUserCredentialsException.cs (hunk 24 above)
Comment (gregory-butler) [reply CredentialsManager/FilesExceptions/DuplicateUserCredentialsException.cs]: I updated it in [493cd86](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/493cd86aa678a20b0c535f9fbea6a18198e61285)

File: CredentialsManager/FilesExceptions/MyFileNotFoundException.cs (hunk 26 above)
Comment (gregory-butler) [reply CredentialsManager/FilesExceptions/MyFileNotFoundException.cs]: I chose to use the long string :)
I updated it in commit [493cd86](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/493cd86aa678a20b0c535f9fbea6a18198e61285)

File: CredentialsManager/Models/Credentials.cs (hunk 30 above)
Comment (gregory-butler) [reply CredentialsManager/Models/Credentials.cs]: At the time you were reviewing my code I was in process of converting the class to a struct as per Kai's suggestion and so I did. I think it is required for classes but not for structs.
The commit is this one [4b30e14](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/4b30e147718194bba663095be2c2d4fa23a90759)

File: CredentialsManager/FilesExceptions/DuplicateUserCredentialsException.cs (hunk 23 above)
Comment (gregory-butler) [reply CredentialsManager/FilesExceptions/DuplicateUserCredentialsException.cs:3]: I'm new to the OOP and namespaces. At my level of experience  I like having everything organized in their own space.

File: CredentialsManager/Files.cs (hunk 21 above)
Comment (gregory-butler) [reply CredentialsManager/Files.cs]: I decided to catch those exceptions because their summaries lead to one conclusion: the file, directory or path can not be found. The 'ReadToEnd()' method's summary for 'IOException' is indeed different from the one of 'StreamReader' object creation.
I refactored the constructor of `MyFileNotFoundException` to include the initial exception thus exposing the `InnerException` property. I hope I understood this correctly.
The commit for this update: [6ba6627](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/6ba6627217ee878b8d508018d4350ad78f7f5a5a)

File: CredentialsManager/Models/Credentials.cs (hunk 31 above)
Comment (gregory-butler) [reply CredentialsManager/Models/Credentials.cs]: I managed to find a proper solution for this (at least I think I do :) )
- I used `OrdinalIgnoreCase` just for `UserName` as the pass should not change case
- I created a `HashCode` object inside `GetHasCode` method, to which I added `UserName` with `OrdinalIgnoreCase`, added the password normally then I convert it and return as the actual hash code int value.

I updated it in this commit: [b7a6dd3](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/b7a6dd3db544d5f8ba4b35d76dc1f478e7268391)

File: CredentialsManager/Manager.cs (hunk 28 above)
Comment (gregory-butler) [reply CredentialsManager/Manager.cs:73]: I lost this one in translation.
Before I refactored with OOP I had this  method
```
private static bool IsUserNameEqual(string[] left, string[] right)
{
  return left[0].Equals(right[0], StringComparison.OrdinalIgnoreCase);
}
```
But now that I implemented `IComparable` I can use `currentCredential.CompareTo(nextCredential).Equals(0)`

gregory-butler: I updated it in commit [712418a](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/712418a54e8d87bd0bfb26656174ddacc93c2d3f)

CredentialsManager/App.cs
gregory-butler: I initially used `Environment.Exit(0)` inside the `Exit` method until commit [0c6e2e8](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/0c6e2e820bd03c5346862ab0599a5f39128cbac8).
Then I read that we shouldn't force an app exit and this how the current iteration appeared.
I thought about it but I couldn't figure another way to do it other than these two.
If there is another way please teach me :)

gregory-butler: I updated it in commit [6426416](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/64264161d395f9f34fae01114b0e3c067e7dae4e)

CredentialsManager/App.cs
gregory-butler: One way I would solve this is to replace `ConsoleKey` with an array or list of the same type and just add both decimal and numpad keys. Calling `Contains` on them should do the trick.
I don't think `ItemId` is redundant despite the fact that I could print the index of the `MenuItem` plus one instead of the id.

gregory-butler: I refactored it to use multiple keys here [156c32e](https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/pull/43/commits/156c32e5012146c3d2066fa777438fc3f0372c7d)

CredentialsManager/Files.cs
gregory-butler: YAGNI got me here 😄 but I won't remove it because I want to practice the suggestions Kai gave me for the tests.

CredentialsManager/Models/Credentials.cs
gregory-butler: I will definitely keep this in mind. Thank you.

CredentialsManager/Models/Menu.cs
gregory-butler: I will definitely give more thought next time I do validation inside a constructor.
Thank you.

gregory-butler:
@westonpace thank you very much again for such a detailed review.

gregory-butler:
@Almantask Thank you very much for the review.
I think I managed to implement the improvements you suggested.

Almantask (approved):
Well done. 🥇

gregory-butler:
Thank you @Almantask and @westonpace you really pushed me with this homework. I learned a lot!

PR #46, opened 2021-04-10
Homework/3

AyameU:
I am a bit confused by the wording of some of the instructions in this week's homework.
Namely:
> If there is no Users.txt file- the application should not throw "UsersNotFoundException.

UsersNotFoundException is a custom exception from what I can see. I'm not clear on why we would create a custom exception, then never throw it?

I assumed it was a typo and did throw the exception at the start of the program. It closes the program before the menu is shown to the user. This may not be the best way to handle the exception, but I don't want the program to continue running when such a critical part is missing.

For my custom exception classes I added the three constructors recommended in the Microsoft docs. https://docs.microsoft.com/en-us/dotnet/standard/exceptions/how-to-create-user-defined-exceptions

### DIFF
--- a/Homework3/Homework3/DuplicateUserCredentialsException.cs
+++ b/Homework3/Homework3/DuplicateUserCredentialsException.cs
@@ -1,0 +1,17 @@
+﻿using System;
+
+class DuplicateUserCredentialsException : Exception
+{
+    public DuplicateUserCredentialsException()
+    {
+    }
+
+    public DuplicateUserCredentialsException(string message) : base(message)
+    {
+    }
+
+    public DuplicateUserCredentialsException(string message, Exception inner)
+        : base(message, inner)
+    {
+    }
+}
--- a/Homework3/Homework3/Program.cs
+++ b/Homework3/Homework3/Program.cs
@@ -139,7 +139,7 @@ namespace Homework3
         {
             string userName;
             string password;
-            bool isDuplicate = false;
+            bool isDuplicate;
 
             do
             {
@@ -148,12 +148,11 @@ namespace Homework3
 
                 try
                 {
-                    isDuplicate = CheckIfDuplicate(filePath, userName);
-                    
+                    isDuplicate = CheckIfDuplicate(filePath, userName); 
                 }
                 catch (DuplicateUserCredentialsException)
                 {
-                    Console.WriteLine(isDuplicate);
+                    isDuplicate = true;
                     Console.WriteLine($"User already exists.{Environment.NewLine}");
                 }
 
@@ -199,13 +198,16 @@ namespace Homework3
 
                 if (isDuplicate)
                 {
-                    reader.Close();
                     break;
-                    throw new DuplicateUserCredentialsException();
                 }
             }
 
             reader.Close();
+
+            if (isDuplicate)
+            {
+                throw new DuplicateUserCredentialsException();
+            }
 
             return isDuplicate;
         }
--- a/Homework3/Homework3/Program.cs
+++ b/Homework3/Homework3/Program.cs
@@ -1,0 +1,179 @@
+﻿using System;
+using System.IO;
+
+namespace Homework3
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            ShowMenu();
+        }
+
+        /// <summary>
+        /// Shows a menu of commands.
+        /// </summary>
+        static void ShowMenu()
+        {
+            string filePath = @"Users.txt";
+            string command;
+
+            while(true)
+            {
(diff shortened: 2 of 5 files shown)


Almantask:
> UsersNotFoundException is a custom exception from what I can see. I'm not clear on why we would create a custom exception, then never throw it?

You're right, thanks! Fixed wiki.

Homework3/Homework3/DuplicateUserCredentialsException.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+class DuplicateUserCredentialsException : Exception
+{
+    public DuplicateUserCredentialsException()
+    {
+    }
+
+    public DuplicateUserCredentialsException(string message) : base(message)
```
Almantask, line 9: Instead, maybe you can say which user is duplicate and just pass a username rather than the full message? You could internally form the message as you want

Homework3/Homework3/Program.cs
```diff
@@ -0,0 +1,238 @@
...
+namespace Homework3
+{
+    /// <summary>
+    /// Program allows users to log in or register new login credentials.
+    /// </summary>
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            string filePath = @"Users.txt";
```
Almantask, line 13: @".." is only needed when you work with `\` (so you don't have to write `\\`). @ is not necessary in this cae.

Homework3/Homework3/Program.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            ShowMenu(filePath);
+        }
+
+        /// <summary>
+        /// Checks that the file exists at the specified path.
+        /// </summary>
+        /// <param name="filePath">The file we are checking.</param>
+        /// <returns>True if the path is valid, or false if invalid.</returns>
+        static void CheckIfFileExists(string filePath)
```
Almantask, line 35: Could be simplified: `FileExists`

Homework3/Homework3/Program.cs
```diff
@@ -0,0 +1,238 @@
...
+            while(true)
+            {
+                Console.WriteLine($"Type a Command:{Environment.NewLine}" 
+                                 + $"Login{Environment.NewLine}"
+                                 + $"Register{Environment.NewLine}"
+                                 + $"Exit{ Environment.NewLine}");
+
+                string command = PromptUserForInput("command");
+
+                if (command.Equals("login", StringComparison.OrdinalIgnoreCase))
```
Almantask, line 66: Those commands could be constants, for clarity and future-proofing.

Homework3/Homework3/Program.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// <returns>
+        /// True if the user name and password match, false if it does not match.
+        /// </returns>
+        static bool CheckIfCredentialsAreValid(string filePath, string userName, string password)
+        {
+            StreamReader reader = new StreamReader(filePath);
+
+            bool isValid = false;
+
+            while (reader.Peek() >= 0)
```
Almantask, line 130: It was stack overflow answer I assume :) 
There is a more simple way: `reader.CanRead`

Homework3/Homework3/Program.cs
```diff
@@ -0,0 +1,238 @@
...
+
+                isValid = words[0].Equals(userName) && words[1].Equals(password);
+
+                if (isValid)
+                {
+                    break;
+                }
+            }
+
+            reader.Close();
```
Almantask, line 143: Prefer to call `Dispose` over `Close` for unamanged resources. Also, prefer to wrap it in a `using(var reader = new StreamReader(path)` block. In your case, closing is the same as disposing.

Homework3/Homework3/Program.cs
```diff
@@ -0,0 +1,238 @@
...
+                }
+                catch (DuplicateUserCredentialsException)
+                {
+                    isDuplicate = true;
+                    Console.WriteLine($"User already exists.{Environment.NewLine}");
+                }
+
+                if (!isDuplicate)
+                {
+                    WriteToDataFile(filePath, userName, password);
```
Almantask, line 176: Won't `SaveCredentials` sound better?

Homework3/Homework3/Program.cs
```diff
@@ -0,0 +1,238 @@
...
+                    Console.WriteLine($"Credentials added.{Environment.NewLine}");
+                    break;
```
Almantask, line 178: Do you really need a break here? Won't the do..while loop stop anyways on this condition?

Almantask (changes requested):

Almantask:
expired

PR #48, opened 2021-04-11
Homework 3

Tailor-AI:
First try on homework 3.

### DIFF
--- a/Homework3/DataAccess.cs
+++ b/Homework3/DataAccess.cs
@@ -1,0 +1,89 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace start
+{
+    public static class DataAccess
+    {
+        private static readonly string fileName = $"Users.txt";
+
+        private static string GetStringsFromFile()
+        {
+            try
+            {
+                using (StreamReader reader = new StreamReader(fileName))
+                {
+                    return reader.ReadToEnd();
+                }
+            }
+            catch (FileNotFoundException ex)
+            {
+                throw new UsersNotFoundException("File Users.txt does not exists.", ex);
+            }
+        }
+
+        public static List<User> GetUsers()
+        {
+            List<User> validUsers = new List<User>();
+
+            string[] validUserStringArray = GetStringsFromFile().Split(Environment.NewLine);
+
+            foreach (var userInfo in validUserStringArray)
+            {
+                string[] userCred = userInfo.Split(",");
+
+                User user = new User();
+                user.Name = userCred[0];
+                user.Password = userCred[1];
+
+                validUsers.Add(user);
+            }
+
+            return validUsers;
+        }
+
+        public static void AddUserToFile(User user)
+        {
+            if (user != null)
+            {
+                string userString = $"{user.Name},{user.Password}";
+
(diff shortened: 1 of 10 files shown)


Homework3/DataAccess.cs
```diff
@@ -0,0 +1,89 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace start
+{
+    public static class DataAccess
```
Almantask: Will the reader know what data access is for?

Homework3/DataAccess.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        private static readonly string fileName = $"Users.txt";
+
+        private static string GetStringsFromFile()
```
Almantask: It's more than just a string, name method after logical data they return, not how it looks like (semantics).

Almantask: Move this method under the ones that consume it.

Homework3/Menu.cs
```diff
@@ -0,0 +1,115 @@
...
+        public static void Run()
+        {
+            Console.WriteLine("Please choose:");
+            Console.WriteLine("1. Login.");
+            Console.WriteLine("2. Register new user.");
+            Console.WriteLine("3. Exit program.");
+
+            switch (ReturnValidUserInput())
+            {
+                case 1:
```
Almantask: could replace magic numbers with constants. In that case- changing the options would be just a matter of changing the constant.

Almantask: IN one place, as opposed to many

Homework3/Menu.cs
```diff
@@ -0,0 +1,115 @@
...
+                string password = Console.ReadLine();
+
+                User newUser = new User { Name = userName, Password = password };
+                DataAccess.AddUserToFile(newUser);
+                Console.WriteLine("User added.");
+
+                Console.WriteLine(Environment.NewLine);
+                Menu.Run();
+            }
+        }
```
Almantask: Try to split console part from logical part. You should consider that, because it makes one dependant on another. If you move on to another way of representing things- you will have to rewrite all, as opposed to just rewriting the representation.

Almantask (changes requested):

Homework3/DataAccess.cs
Tailor-AI: Renamed this class to UserFileLogic.

Homework3/DataAccess.cs
Tailor-AI: Renamed to GetUserCredentialsFromFile and moved down.

Homework3/Menu.cs
Tailor-AI: I made constants instead and used their names. Hopefully I understood this correctly?

Homework3/Menu.cs
Tailor-AI: I moved all Console command to a seperate ConsoleUI class and used this through a interface so I can later add eg. WinFormsUI etc.

Homework3/Menu.cs
Almantask: Yup, all good 👍

Almantask:
Looks better :)

PR #52, opened 2021-04-22
Homework/4

Laimonukas:
Added:
Recipe validation(empty,cooking units,values);
OpenFileDialog filters only .recipe files;
Cooking units get converted to proper SI units, or get scaled up;
SI units get converted back to cooking units;
ml scales to l when >100ml;

### DIFF
--- a/RecipeApp.Client/Form1.cs
+++ b/RecipeApp.Client/Form1.cs
@@ -16,6 +16,7 @@ namespace RecipeApp.Client
         public RecipeApp()
         {
             InitializeComponent();
+            RecipeConverter.InitialSetup();
         }
 
         private void LoadButton_Click(object sender, EventArgs e)
@@ -28,6 +29,12 @@ namespace RecipeApp.Client
             // Prepare for browsing files
             using (var dialog = new OpenFileDialog())
             {
+                // Adding restrictions and initial dialog settings:
+                dialog.InitialDirectory = Environment.CurrentDirectory;
+                dialog.Title = "Browse Recipe files";
+                dialog.DefaultExt = "recipe";
+                dialog.Filter = "recipe files (*.recipe)|*.recipe";
+
                 // Browse files... if file was confirmed
                 if (dialog.ShowDialog() == DialogResult.OK)
                 {
@@ -35,11 +42,21 @@ namespace RecipeApp.Client
                     var stream = dialog.OpenFile();
                     using (var reader = new StreamReader(stream))
                     {
-                        return reader.ReadToEnd();
+                        
+                        string recipeText = reader.ReadToEnd();
+                        string validationText = RecipeConverter.RecipeValidation(recipeText);
+
+                        // File validation:
+                        if (validationText != "")
+                        {
+                            MessageBox.Show(validationText);
+                        }
+                        else {
+                            return recipeText;
+                        }
                     }
                 }
             }
-
             // Happens if canceled.
             return string.Empty;
         }
@@ -48,7 +65,16 @@ namespace RecipeApp.Client
         {
             var isConvertToSiUnits = SiUnitsRadioButton.Checked;
             var input = SourceRichTextBox.Text;
-            ResultRichTextBox.Text = RecipeConverter.ConvertRecipe(input, isConvertToSiUnits);
+            // Check the recipe
+            string validationText = RecipeConverter.RecipeValidation(input);
+            if (validationText == "")
+            {
+                ResultRichTextBox.Text = RecipeConverter.ConvertRecipe(input, isConvertToSiUnits);
+            }
+            else 
+            {
+                MessageBox.Show(validationText);
+            }
         }
     }
 }
--- a/RecipeApp.Core/RecipeConverter.cs
+++ b/RecipeApp.Core/RecipeConverter.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Globalization;
+using System.Collections.Generic;
+using System.Linq;
 using System.IO;
 using System.Net;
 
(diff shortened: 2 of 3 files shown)


RecipeApp.Client/Form1.cs
```diff
@@ -28,18 +29,34 @@ private static string ReadContentsFromFileBrowser()
             // Prepare for browsing files
             using (var dialog = new OpenFileDialog())
             {
+                // Adding restrictions and initial dialog settings:
+                dialog.InitialDirectory = Environment.CurrentDirectory;
+                dialog.Title = "Browse Recipe files";
+                dialog.DefaultExt = "recipe";
+                dialog.Filter = "recipe files (*.recipe)|*.recipe";
```
Almantask, line 36: Move to function: `SetupDialog`

RecipeApp.Client/Form1.cs
```diff
@@ -28,18 +29,34 @@ private static string ReadContentsFromFileBrowser()
...
+
+                        // File validation:
+                        if (validationText != "")
+                        {
+                            MessageBox.Show(validationText);
+                        }
+                        else {
+                            return recipeText;
+                        }
                     }
```
Almantask, line 57: Move to function: `LoadRecipe`

RecipeApp.Client/Form1.cs
```diff
@@ -48,7 +65,16 @@ private void ConvertButton_Click(object sender, EventArgs e)
         {
             var isConvertToSiUnits = SiUnitsRadioButton.Checked;
             var input = SourceRichTextBox.Text;
-            ResultRichTextBox.Text = RecipeConverter.ConvertRecipe(input, isConvertToSiUnits);
+            // Check the recipe
+            string validationText = RecipeConverter.RecipeValidation(input);
+            if (validationText == "")
```
Almantask, line 70: Prefer to use `string.Empty` over `""`

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -1,14 +1,176 @@
...
 
 namespace RecipeApp
 {
     public class RecipeConverter
     {
-        static double[] multipliers = {240, 14.79, 4.93};
-        static string[] units = {"cup", "tablespoon", "teaspoon"};
+        static double[] multipliers = {240,3790,29.57,470,950,14.79,4.93};
+        static string[] units = {"cup","gallon","fluid ounce","pint","quart","tablespoon","teaspoon"};
+        static List<string> doubleUnits = new List<string>();
```
Almantask, line 14: What do you mean by `doubleUnits`?

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -1,14 +1,176 @@
...
+
+        // Entry method to get some initial setup for Lists
+        public static void InitialSetup() {
```
Almantask, line 17: Name could be simplified: `Initialize`

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -1,14 +1,176 @@
...
+            PopulateDoubleUnitsList();
+            SortCookingUnitsByMultiplier();
+        }
+
+        public static string RecipeValidation(string recipeText) {
```
Almantask, line 22: Methods do things, therefore they should be named after verbs rather than nouns. rename to: `ValidateRecipe`

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -1,14 +1,176 @@
...
+                    }
+                }
+                if (sorted)
+                {
+                    return;
+                }
+            }
+        }
+
+        // Validation method to check if string contains any of the cooking units
```
Almantask, line 70: If you put trippled slash (`///`) it will create a method documentation with a summary comment. Prefer to use that, because it will allow reading the what the method does when you hover your mouse over it.

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -1,14 +1,176 @@
...
+                {
+                    containsUnits = true;
+                }
+            }
+            if (wordList.Contains("ml") || wordList.Contains("l"))
+            {
+                containsUnits = true;
+            }
+            return containsUnits;
+        }
```
Almantask, line 86: Is there much meaning from this method? The idea behind validating if it is a valid cooking measurement was that there should always be a number, followed with some unit of measurement.

For example "10 ml" is valid. "10 tsp" is valid as well, but "x ml" is not.

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -1,14 +1,176 @@
...
+
+        // Check method to ensure that each unit as some value number before it
+        // returns -1 if ok, else returns index of bad cooking unit
+        public static int ReturnFirstFaultyCookingUnitIndex(string textToCheck) {
```
Almantask, line 90: Avoid using a word `Return` in a method name. It sounds forced. Try to be more natural, for example:
`FindFirstFaultyCookingUnitIndex`

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -1,14 +1,176 @@
...
+            List<string> words = new List<string>(textToCheck.ToLower().Split(' ', '\n'));
```
Almantask, line 91: I'd split this into two lines- when you do 2 very different things (create a list and split lowercased text) it makes it significantly more difficult to read.

Almantask: The analogy is this; which reads better for a shopping list:
- 10 eggs
- 5 apples
- 1 kiwi
or
10 eggs, 5 apples, 1 kiwi
?

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -1,14 +1,176 @@
...
+            textToCheck = FixDoubleWordedCookingUnits(textToCheck);
+            
+            // looking for cooking units
+            for (int i = 0; i < units.Length; i++)
+            {
+                string lookFor = units[i].Replace(" ", "_").ToLower();
+                int foundUnitIndex = words.FindIndex(x => x.Contains(lookFor));
+                if (foundUnitIndex != -1)
+                {
+                    if (foundUnitIndex != 0)
```
Almantask, line 101: state, rather than negate.

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -1,14 +1,176 @@
...
+                    else
+                    {
+                        return 0;
+                    }
+                }
+                else
+                {
+                    continue;
+                }
+            }
```
Almantask, line 121: Refactor:
```cs
for (int i = 0; i < units.Length; i++)
{
    string lookFor = units[i].Replace(" ", "_").ToLower();
    int foundUnitIndex = words.FindIndex(x => x.Contains(lookFor));

    if (foundUnitIndex == -1) continue;
    if (foundUnitIndex == 0) return 0;

    var isNumber = ParseDouble(words[foundUnitIndex - 1], out _);
    if (!isNumber) return foundUnitIndex;

    words.RemoveRange(foundUnitIndex - 1, 2);
    i--;
}
```
Note: `_` is a discard operator- used when you don't care about the out- you just want the normal return, like in your case- just to see if Tryparse returns true or false. More info: https://docs.microsoft.com/en-us/dotnet/csharp/discards

Almantask (changes requested):
Overall doesn't look too bad :) Well done 👍

PR #53, opened 2021-04-24
Homework 4

Tailor-AI:
Added the requirements to RecipeApp solution from lesson 4.

### DIFF
--- a/RecipeApp.Client/Form1.cs
+++ b/RecipeApp.Client/Form1.cs
@@ -1,18 +1,13 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.ComponentModel;
-using System.Data;
-using System.Drawing;
 using System.IO;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using System.Windows.Forms;
+
 
 namespace RecipeApp.Client
 {
     public partial class RecipeApp : Form
     {
+
         public RecipeApp()
         {
             InitializeComponent();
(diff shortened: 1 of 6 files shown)


RecipeApp.Client/Form1.cs
```diff
@@ -25,30 +20,44 @@ private void LoadButton_Click(object sender, EventArgs e)
...
+                        using (var reader = new StreamReader(stream))
+                        {
+                            return reader.ReadToEnd();
+                        }
+                    }
+                    else
+                    {
+                        MessageBox.Show("Invalid recipe file. Must end with .recipe");
                     }
                 }
```
Almantask: Dialog has an option to filter on initial path and extensions, try to make use of that 😄

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -76,16 +57,43 @@ static void StandardiseCookingUnit(int index, string[] words)
...
+        {
+            if (amountMl >= 100)
+            {
+                return (amountMl / 1000).ToString("F2", CultureInfo.InvariantCulture);
+            }
+            else
+            {
+                return amountMl.ToString("F2", CultureInfo.InvariantCulture);
+            }
+        }
```
Almantask: can be simplified:
```cs
        private static string GetAmountInSiUnits(double amountMl)
        {
            if(amountMl >= 100)
            {
                amountMl = amountMl / 100;
            }
            return amountMl.ToString("F2", CultureInfo.InvariantCulture);
        }
```

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -94,29 +102,59 @@ static double FindMultiplier(string cookingUnit)
...
+
+            if (!containsCookingOrSiUnits)
+            {
+                throw new InvalidRecipeException("No SI units in recipe.");
+            }
+
+            return string.Join(" ", words);
+        }
+
+        private static void StandardiseSiUnit(int index, string[] words)
```
Almantask: Instead, make this method return  a bool so you won't need to use a field but rather have it controlled in local scope. That confused me a bit.

Almantask (changes requested):
Try to slightly be more clear with error messages and please use dialog-native validation.

RecipeApp.Client/Form1.cs
Tailor-AI: Implemented and removed the previous check.

RecipeApp.Core/RecipeConverter.cs
Tailor-AI: Ahh. Didnt think of that. Thx!

RecipeApp.Core/RecipeConverter.cs
Tailor-AI: Have removed the field, made function return bool and moved the checking of different units into respective function.

Almantask:
Looks good now :)

PR #55, opened 2021-04-25
Homework/4

cjahermanns:


RecipeApp.Core/RecipeConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+        //        if (multiplier == -1) continue;
+
+        //        ParseDouble(words[index - 1], out amount);
+
+        //        words[index] = GetClosestCookingUnit(amount);
+        //        words[index - 1] = (amount * multiplier / Conversion.CookingUnits[words[index]]).ToString("F2");
+        //    }
+
+        //    return string.Join(" ", words);
+        //}
```
Almantask: Don't leave commented out code.

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+
+        public static string MakeRecipePretty(string words)
```
Almantask: The name of a function is not clear. Instead of pretty, state what you do with a recipe. If it's not possible to express that with a reasonably short method name, use a comment.

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+                if ( tempUnitName == unitName || tempUnitName+ "s" == unitName)
+                {
+                    unitType = baseUnitType;
+                    break;
+                }
+            }
+            if (unitType is null) return null;
+
+            BaseUnit unit = null;
+            if (ParseDouble(words[idx - 1], out double amount))
```
Almantask: It's more readable if you move this to a separate line.

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+            foreach (var baseUnitType in Assembly.GetExecutingAssembly().GetTypes().Where(t => (t.BaseType?.Name ?? "") == "BaseUnit"))
+            {
+                var tempUnit = Assembly.GetExecutingAssembly().CreateInstance(baseUnitType.FullName) as BaseUnit;
+                var tempUnitName = baseUnitType.GetProperties().FirstOrDefault(f =>f.Name == "Name")?.GetValue(tempUnit).ToString() ?? "";
+                if ( tempUnitName == unitName || tempUnitName+ "s" == unitName)
+                {
+                    unitType = baseUnitType;
+                    break;
+                }
+            }
```
Almantask: Avoid doing reflection. Consider redoing these bits and simplifying it.

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+                foreach (var baseUnitType in Assembly.GetExecutingAssembly().GetTypes().Where(t => (t.BaseType?.Name ?? "") == "BaseUnit"))
+                {
+                    var tempUnit = Assembly.GetExecutingAssembly().CreateInstance(baseUnitType.FullName) as BaseUnit;
+                    var tempUnitName = baseUnitType.GetProperties().FirstOrDefault(f => f.Name == "Name")?.GetValue(tempUnit).ToString() ?? "";
+                    if (tempUnitName == newUnitName || tempUnitName + "s" == newUnitName)
+                    {
+                        unitType = baseUnitType;
+                        break;
+                    }
+                }
```
Almantask: Also, this is a duplicate part of code.

RecipeApp.Core/RecipeConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+
+            if (amountMl > 100)
+            {
+                words[index] = "l";
+                words[index - 1] = (amountMl / 100).ToString("F2", CultureInfo.InvariantCulture);
+                return;
+            }
+
+            words[index] = "ml";
+            words[index - 1] = amountMl.ToString("F2", CultureInfo.InvariantCulture);
```
Almantask: Converting a number to a consistent format should go to a static function of its own, so that we consistently apply the same format throughout.

RecipeApp.Core/Units/BaseUnit.cs
```diff
@@ -0,0 +1,35 @@
...
+
+            return output;
+        }
+
+        public override string ToString()
+        {
+            return $"{Amount:F2} {Name}";
+        }
+    }
+}
```
Almantask: It can be a "normal" class and just create a bunch of instances of it, if you want to go down the OOP route.

RecipeApp.Core/Units/Cooking/Gallon.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace RecipeApp.Core.Units.Cooking
+{
+    public class Gallon : BaseUnit
+    {
+        public override string Name => "gallon";
```
Almantask: Instead of hardcoding the name and creating a ton of classes, create multiple objects.

Almantask: The unit itself can be an enum if you want the OOPines.

Almantask (changes requested):
Try to keep things more simple.

RecipeApp.Core/RecipeConverter.cs
cjahermanns: Would you say only leave main / master branch clean of comments (and PRs actually) or would you also suggest removing those lines in develop / wip branches where you try different method implementations?

RecipeApp.Core/RecipeConverter.cs
cjahermanns: What's your rule of thumb how long a method name should be?

RecipeApp.Core/RecipeConverter.cs
cjahermanns: extracted a method

RecipeApp.Core/RecipeConverter.cs
cjahermanns: removed it and tried a different approach...

RecipeApp.Core/RecipeConverter.cs
cjahermanns: reformatted the whole thing ;D

RecipeApp.Core/RecipeConverter.cs
cjahermanns: Now it's gone.

RecipeApp.Core/Units/BaseUnit.cs
cjahermanns: Went with that suggestion!

RecipeApp.Core/Units/Cooking/Gallon.cs
cjahermanns: I used this approach to be able to use reflection and multiword units... Now I use only one `Unit`class

cjahermanns:
Can I add the logging hw in here or shall it be in another PR?

RecipeApp.Core/RecipeConverter.cs
Almantask: I would say upon pushing- they should be clean. Sorry for a late response.

RecipeApp.Core/RecipeConverter.cs
Almantask: The length is irrelevant. But it should do one thing. So, if you ever consider splitting a method- split it. Don't force the reader to read through all the code (regardless how big it is) if all you want to understand is the big picture. If you want to know the details, you will go deeper into private methods.

Method doing 1 thing- means 1 level of abstraction. So for example if you call a method SendEmail and next to that you run a loop- loop is a low level and thus should be moved to a method of its own.

Almantask (approved):
Looks good :)

RecipeApp.Core/RecipeConverter.cs
cjahermanns: Alright ty!

RecipeApp.Core/RecipeConverter.cs
cjahermanns: Ok , got it ^^

PR #56, opened 2021-04-29
Homework 5

cjahermanns:


People.API/Controllers/PeopleController.cs
```diff
@@ -0,0 +1,120 @@
...
+    {
+        private static Dictionary<int, Person> _persons = new Dictionary<int, Person>();
+
+        [HttpPost]
+        public IActionResult AddPerson([FromBody] Person value)
+        {
+            while (!_persons.TryAdd(value.Id, value))
+            {
+                value.Id++;
+            }
```
Almantask: It works, but are those the requirements?

People.API/Controllers/PeopleController.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            return Created($"/people/{value.Id}", value);
+        }
+
+        [HttpPut("{id:int}/pets")]
+        public IActionResult AddPetToPerson(int id, [FromBody] Pet value)
+        {
+            if (_persons.TryGetValue(id, out var owner))
```
Almantask: Move this to a separate line for better readability.
Validate first, meaning which if it's invalid- return not found and only then continue the rest of logic. That way, you eliminate the scope you don't need to care about fast, continuing with the essential "work as intended" behavior.

People.API/Controllers/PeopleController.cs
```diff
@@ -0,0 +1,120 @@
...
+            {
+                owner.Pets.Add(value);
+
+                return Ok(owner);
+            }
+
+            return NotFound($"The person with id {id} could not be found.");
+        }
+
+        [HttpPatch("{id:int}")]
```
Almantask: You don't have to specify it's an it. I think it's a default based on the argument.

People.API/Controllers/PeopleController.cs
```diff
@@ -0,0 +1,120 @@
...
+        public IActionResult ChangePerson(int id, [FromBody] Person value)
+        {
+            if (id != value.Id) return BadRequest();
```
Almantask: You should always specify a reason for a bad request.

People.API/Controllers/PeopleController.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        [HttpDelete("{id:int}")]
+        public IActionResult DeletePerson(int id)
+        {
+            if (_persons.ContainsKey(id))
+            {
+                _persons.Remove(id);
+
+                return NoContent();
```
Almantask: NoContent refers that you were trying to get something and then there was nothing of that something. A successful deletion could say `Ok("Person with id X was deleted");` or just an `Ok();`

People.API/Controllers/PeopleController.cs
```diff
@@ -0,0 +1,120 @@
...
+            if (_persons.TryGetValue(id, out var person))
+            {
+                return Ok(person);
+            }
+
+            return NotFound($"The person with id {id} could not be found.");
+        }
+
+        [HttpGet]
+        public IActionResult GetPersons()
```
Almantask: People not persons.

People.API/Controllers/PeopleController.cs
```diff
@@ -0,0 +1,120 @@
...
+            return NotFound($"The person with id {id} could not be found.");
+        }
+
+        [HttpDelete("{id:int}/pets/{petName}")]
+        public IActionResult RemovePet(int id, string petName)
+        {
+            if (_persons.TryGetValue(id, out var person))
+            {
+                if (person.Pets.Exists(p => p.Name.Equals(petName, System.StringComparison.OrdinalIgnoreCase)))
+                {
```
Almantask: This is not needed, because the other part will simply do nothing anyways if there are no pets of that person.

People.API/Models/Person.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace People.API.Models
+{
+    public class Person
+    {
+        public int Age { get; set; }
+        public int Id { get; set; }
+        public string Name { get; set; }
+        public List<Pet> Pets { get; set; } = new();
+        public int SpouseId { get; set; } = -1;
```
Almantask: use a constant to specify the meaning of `-1`

Almantask (changes requested):

People.API/Models/Person.cs
cjahermanns: Good idea! Never thought about using consts to declare special meanings xD

People.API/Controllers/PeopleController.cs
cjahermanns: Nope... changed it to either `Created()` or `BadRequest()`. Is there a more appropriate result?

People.API/Controllers/PeopleController.cs
cjahermanns: Ok, I removed the `int`s

People.API/Controllers/PeopleController.cs
cjahermanns: Did so in the add method, removed this one since I don't think it's necessary...

People.API/Controllers/PeopleController.cs
cjahermanns: You are right, Got distracted from all the `singular`s suggestions intellisense gives...

People.API/Controllers/PeopleController.cs
cjahermanns: Ok, changed it to respond with either `Ok()` if the pet was removed, or `NotFound()` if there was no such pet.

People.API/Controllers/PeopleController.cs
cjahermanns: Got rid of it ^^

People.API/Controllers/PeopleController.cs
cjahermanns: Well I now use a `FindPerson` method but is there a way to remove the duplicate `if (person == null) return NotFound()` stuff? I tried something like 
```cs
private IActionResult WorkWithPerson(int id, Action<Person> action)
{
  if (_people.TryGetValue(id, out var person))
  {
     action(person);
     return Ok(person);
  }
  return NotFound($"The person with id {id} could not be found.");
}
```
but I didn't like how limited I was with that approach...

Almantask (approved):
Looks good

PR #57, opened 2021-05-01
Homework 5

Tailor-AI:
A lot of new stuff to take in. Dont know if I understood the marry task correct.

### DIFF
--- a/ShoppingListApi/Controllers/PersonController.cs
+++ b/ShoppingListApi/Controllers/PersonController.cs
@@ -100,10 +100,8 @@ namespace PersonPetsApi.Controllers
                 return NotFound($"Person by id {id} was not found.");
             }
 
-            oldPerson.Pets = person.Pets;
-            oldPerson.Name = person.Name;
-            oldPerson.Age = person.Age;
-            oldPerson.SpouseId = person.SpouseId;
+            oldPerson.UpdatePerson(person);
+
             return Ok();
         }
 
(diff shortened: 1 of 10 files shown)


ShoppingListApi/Controllers/PersonController.cs
```diff
@@ -0,0 +1,155 @@
...
+
+            if (oldPerson == null)
+            {
+                return NotFound($"Person by id {id} was not found.");
+            }
+
+            oldPerson.Pets = person.Pets;
+            oldPerson.Name = person.Name;
+            oldPerson.Age = person.Age;
+            oldPerson.SpouseId = person.SpouseId;
```
Almantask: Out of place. Move this to a person class.

ShoppingListApi/Controllers/PersonController.cs
```diff
@@ -0,0 +1,155 @@
...
+            }
+
+            return Ok();
+        }
+
+        private void Marry(Person person1, Person person2)
+        {
+            person1.SpouseId = person2.Id;
+            person2.SpouseId = person1.Id;
+        }
```
Almantask: Move this to a person class.

Almantask (changes requested):

ShoppingListApi/Controllers/PersonController.cs
Tailor-AI: Moved this logic to Person class.

ShoppingListApi/Controllers/PersonController.cs
Tailor-AI: Moved to Personclass.

Almantask (approved):

PR #58, opened 2021-05-02
Implemented homework

Almantask:
Is just a test.


Almantask (commented):
This is not compliant, missing XYZ

PR #59, opened 2021-05-02
Cleanup

Almantask:


ShoppingListApi/ShoppingListApi/Item.cs
```diff
@@ -1,9 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
```
Almantask: Why was this removed? Please undo

Almantask: Or cleanup in all places.

Almantask: Removed using statements in all places, consistentlky

PR #60, opened 2021-05-03
homework/3

ghost:
Please let me know if anything needs to be changed


Authentication/Authentication/Program.cs
```diff
@@ -0,0 +1,139 @@
...
+                else if (choice.Equals("Register", StringComparison.OrdinalIgnoreCase))
+                {
+                    Register(path);
+                    Console.WriteLine();
+                }
+                else
+                {
+                    Console.WriteLine("Invalid choice: {0}", choice);
+                    Console.WriteLine();
+                }
```
Almantask, line 47: Note the difference between: ` CheckDuplicateUsers(path);` and the while loop. While loop requires you to go deep into details and read through the whole function just to figure out what the intention was. Instead, use a meaningful function name to clarify the intention. So split this function (while) into another.

Authentication/Authentication/Program.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            var path = @"./Data/Users.txt";
+            string choice = "0";
+
+            CheckDuplicateUsers(path);
+
+            while (!choice.Equals("Exit", StringComparison.OrdinalIgnoreCase))
+            {
+                choice = GetUserChoice();
+                if (choice.Equals("Login", StringComparison.OrdinalIgnoreCase))
```
Almantask, line 23: Instead of repeating the same capitalization ignoring compare, make choice lowercase and then you will even be able to use a `switch` statement.

Authentication/Authentication/Program.cs
```diff
@@ -0,0 +1,139 @@
...
+                    }
+                }
+            }
+
+            using var sw = new StreamWriter(path, true);
+            sw.Write(userRegisterString + Environment.NewLine);
+
+            // added these line so that the userRegisterString gets appended to the Users.txt present
+            // in the project folder as the during runtime it was only getting appended
+            // to the bin/dotnetcoreapp/debug/Users.txt file
```
Almantask, line 117: Why is that not ok?

Authentication/Authentication/UsersNotFoundException.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Authentication
+{
+    class UsersNotFoundException : FileNotFoundException
+    {
+        public UsersNotFoundException(string reason) : base(reason)
```
Almantask, line 10: Instead of inputing a reason every time, simply supply a user id which was tried to be found, but wasn't.

Almantask (changes requested):

Authentication/Authentication/Program.cs
ghost: If I was adding users during the execution of the program it was only adding the register info in the Users.txt present in the debug folder (which is generated because of building the project) but when the project is ran again the unchanged Users.txt file present in project folder would replace the file present in the debug folder. So the new users who were registered during the execution of the program would not be recognised by Login( ) on the second run, so to combat this problem I added lines 119 and 120 which during the execution changed the Users.txt present inside the projects folder. So if you registered by running the programming and then exited the program and re-ran the program the users credentials used to registered can be used to Login( ) now.

Almantask: > when the project is ran again the unchanged Users.txt file present in project folder would replace the file present in the debug folder

You can select a property for "copy file" to "replace if newer". Your changed files will be newer, so they would not be replaced anymore.

ghost: Oh okay, I didn't knew that, I'll make the changes.

Authentication/Authentication/Program.cs
ghost: Ok, I thought that the menu system should accept all types of capitalisation disregarding the capitalisation (whether a user users Login or login or LOGIN or LoGiN) just as a fail safe mechanism.

Authentication/Authentication/Program.cs
ghost: Changed it, Now it is working as intended.

Authentication/Authentication/UsersNotFoundException.cs
ghost: This exception is thrown when the "Users.txt" file is not found, according to the Homework question -> errors, I didn't implement any system that would throw an exception if the login credentails were not found. If the login credentials are not found the program just ignores and prompt the user for choice again.

PR #61, opened 2021-05-04
Homework/5

Laimonukas:
Added functions:
Add person (POST),
Get all persons (GET),
Get person by ID (GET),
Adding pets to person (PUT),
Remove person (DELETE),
Remove pet (PUT),
Update person (PATCH),
Marry persons (PUT)

### DIFF
--- a/Homework5/Controllers/PeopleApiController.cs
+++ b/Homework5/Controllers/PeopleApiController.cs
@@ -39,41 +39,41 @@ namespace PeopleApi.Controllers
         [HttpGet("{id}")]
         public IActionResult GetPersonById(int id)
         {
-            Person tempPersonObj = FindPersonById(id);
+            Person personObj = FindPersonById(id);
 
-            if (tempPersonObj == null)
+            if (personObj == null)
             {
                 return NotFound($"Person by id {id} was not found.");
             }
 
-            return Ok(tempPersonObj);
+            return Ok(personObj);
         }
 
         [HttpPut("{id}")]
         public IActionResult AddPetToPersonChosenById(int id, Pet petObj)
         {
-            Person tempPersonObj = FindPersonById(id);
-            if (tempPersonObj == null)
+            Person personObj = FindPersonById(id);
+            if (personObj == null)
             {
                 return NotFound($"Person by id {id} was not found.");
             }
 
-            tempPersonObj.Pets.Add(petObj);
+            personObj.Pets.Add(petObj);
 
-            return Ok(tempPersonObj);
+            return Ok(personObj);
         }
 
         [HttpDelete("{id}")]
         public IActionResult RemovePerson(int id)
         {
-            Person tempPersonObj= FindPersonById(id);
+            Person personObj= FindPersonById(id);
 
-            if (tempPersonObj == null)
+            if (personObj == null)
             {
                 return NotFound($"Shopping list by id {id} was not found.");
             }
 
-            personList.Remove(tempPersonObj);
+            personList.Remove(personObj);
 
             return Ok();
         }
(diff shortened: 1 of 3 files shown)


Homework5/Controllers/PeopleApiController.cs
```diff
@@ -0,0 +1,198 @@
...
+            }
+
+            Pet tempPetObj = FindPetByName(tempPersonObj.Pets, petName);
+            if (tempPetObj == null)
+            {
+                return NotFound($"Pet by name of {petName} was not found.");
+            }
+
+            RemovePetObjectFromPersonObject(tempPersonObj, tempPetObj);
+            tempPersonObj = FindPersonById(id);
```
Almantask: Why do you need to find a person again?

Homework5/Controllers/PeopleApiController.cs
```diff
@@ -0,0 +1,198 @@
...
+            personList.Remove(tempPersonObj);
+
+            return Ok();
+        }
+
+        [HttpPut("{id}/{petName}")]
+        public IActionResult RemovePetFromPerson(int id, string petName)
+        {
+            Person tempPersonObj = FindPersonById(id);
+            if (tempPersonObj == null)
```
Almantask: temp and Obj in the name are not necessary, they don't add value nor clarify intention. Usually we use those as intermediate values, but in this case they are the actuals.

Homework5/Controllers/PeopleApiController.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        [HttpPut("{id}/{spouseId}/marry")]
+        public IActionResult MarryPersons(int id,int spouseId) 
+        {
+            if (FindPersonById(id) == null || FindPersonById(spouseId) == null)
+            {
+                return NotFound($"Person by id {id} was not found.");
+            }
+            ChangeSpouseIds(id, spouseId);
+
```
Almantask, line 121: Move this to person class.

Homework5/Controllers/PeopleApiController.cs
```diff
@@ -0,0 +1,198 @@
...
+            return Ok();
+        }
+
+        /// <summary>
+        /// Returns found Pet object when provided a List of Pet object and search name
+        /// </summary>
+        /// <param name="petList">List to search in</param>
+        /// <param name="petName">Name to search for</param>
+        /// <returns></returns>
```
Almantask, line 130: Instead of focusing on documenting private methods- do this for your public ones.

Homework5/Controllers/PeopleApiController.cs
```diff
@@ -0,0 +1,198 @@
...
+            if (personReference != null)
+            {
+                for (int i = 0; i < personList.Count; i++)
+                {
+                    if (personList[i] == personReference)
+                    {
+                        personList[i] = newPersonObj;
+                    }
+                }
+            }
```
Almantask: This does not really update a person, but rather swap the element of a person list. Not quite what was asked.
Also, don't rely on matching references, simply set all properties of the old object to properties of the new object.

Almantask: To find the person you need you will need an equality by id.

Almantask: Could also be moved to a person class, because it can manage itself just fine without relying on anything. Controller should be as thin as possible.

Homework5/Controllers/PeopleApiController.cs
Laimonukas: I wanted to return an actual changed value, and first time finding the object was to get the reference. This is actually unneeded.

Homework5/Controllers/PeopleApiController.cs
Laimonukas: I don't understand why this needs to be moved to Person class, as this function changes two different objects so it still needs to query through available object list to get correct spouses. Also it would require a list of objects be passed to when in person class. Unless I'm missing something obvious

Almantask: You're right. Another person had Marry function or similar, which changed ids of already found person and existing person to update them accordingly. I think you could spli this function into finding the person and then updating the ids on both through that method. Updating two ids is very simple and does not reduce complexity, but method called `Mary` clarifies the intention of two id changes for spouses.

Homework5/Controllers/PeopleApiController.cs
```diff
@@ -39,63 +39,63 @@ public IActionResult GetAllPersons()
...
-            if (tempPersonObj == null)
+            Person personObj = FindPersonById(id);
+            if (personObj == null)
             {
                 return NotFound($"Person by id {id} was not found.");
             }
 
-            Pet tempPetObj = FindPetByName(tempPersonObj.Pets, petName);
-            if (tempPetObj == null)
+            Pet petObj = FindPetByName(personObj.Pets, petName);
```
Almantask, line 90: `Obj` is noise. It doesn't add to clarity.

Almantask: So it can be removed from a variable name.

Homework5/Person.cs
```diff
@@ -12,5 +12,18 @@ public class Person
...
+        public void UpdatePerson(Person newDetails) {
+            if (Id != newDetails.Id) Id = newDetails.Id;
+
+            if (SpouseId != newDetails.SpouseId) SpouseId = newDetails.SpouseId;
+
+            if (Name != newDetails.Name) Name = newDetails.Name;
+
+            if (Age != newDetails.Age) Age = newDetails.Age;
+
+            if (Pets != newDetails.Pets) Pets = newDetails.Pets;
```
Almantask, line 26: You can make it more simple- just assign the value. If it is not the same- it will be the same. If it was the same- nothing will change when you set the same value to the same value.

Almantask: So remove the ifs.

Almantask (changes requested):

PR #63, opened 2021-05-07
Homework/5

aus10code:
3 relevant files are

Person.cs
Pet.cs
PersonController.cs

### DIFF
--- a/Person/People/Controllers/PersonController.cs
+++ b/Person/People/Controllers/PersonController.cs
@@ -1,0 +1,141 @@
+﻿using System.Collections.Generic;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Person.Controllers
+{
+    [ApiController]
+    [Route("api/[controller]")]
+    public class PersonController : ControllerBase
+    {
+        private static List<Person> _personList = new List<Person>();
+        
+        [HttpPost]
+        public IActionResult Create(Person person)
+        {
+            _personList.Add(person);
+
+            return Ok(person);
+        }
+
+        [HttpPatch("{id}")]
+
+        public IActionResult AddPet(Pet pet, int id)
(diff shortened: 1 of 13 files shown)


Person/People/Controllers/PersonController.cs
```diff
@@ -0,0 +1,141 @@
...
+            if (_person == null)
+            {
+                return NotFound($"person with id:{id} not found");
+            }
+
+            _person.Age = person.Age;
+            _person.Gender = person.Gender;
+            _person.Name = person.Name;
+            _person.PetList = person.PetList;
+            _person.SpouseId = person.SpouseId;
```
Almantask, line 50: move this to person class.

Person/People/Controllers/PersonController.cs
```diff
@@ -0,0 +1,141 @@
...
+            var person1 = GetPerson(id1);
+            var person2 = GetPerson(id2);
+
+            if (person1 == null || person2 == null)
+            {
+                return NotFound($"One of the people was not found");
+            }
+
+            person1.SpouseId = person2.Id;
+            person2.SpouseId = person1.Id;
```
Almantask, line 104: could move this to person class.

Person/People/Controllers/PersonController.cs
```diff
@@ -0,0 +1,141 @@
...
+            foreach (var pet in person.PetList)
+            {
+                if (pet.Name == petName)
+                {
+                    return pet;
+                }
+            }
+
+            return null;
+        }
```
Almantask, line 120: could be moved to person class.

Almantask (changes requested):
Keep controller logic as thin as possible and if an object can manage itself without involving other components- it should.

PR #67, opened 2021-05-23
homework/3

ghost:
Sorry for late responses, my previous PR was deleted/marked expired, because of inactivity (i think), I have my finals going on i.e. why it took so long for me to respond. I made the changed requested in the deleted/marked expired PR in this new PR


Authentication/Authentication/DuplicateUserCredentialsException.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Authentication
+{
+    class DuplicateUserCredentialsException : Exception
+    {
+        public DuplicateUserCredentialsException()
+        {
+            Console.WriteLine("There are duplicate entries in the file");
```
Almantask, line 11: Don't do this- in some environments we won't be able to see the exception message.
Instead, use this: ` public DuplicateUserCredentialsException(): base("There are duplicate entries in the file"){}`

Almantask: In fact, specify the duplicate entry.

Authentication/Authentication/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+                    }
+                    finally
+                    {
+                        Console.WriteLine();
+                    }
+                }
+                else if (choice.Equals("Register", StringComparison.OrdinalIgnoreCase))
+                {
+                    Register(path);
+                    Console.WriteLine();
```
Almantask, line 48: All parts end with `Console.WriteLine()`- consider moving outside of the if statement.

Authentication/Authentication/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+                        Login(path);
+                    }
+                    catch (FileNotFoundException exception)
+                    {
+                        throw new UsersNotFoundException(exception.FileName);
+                    }
+                    finally
+                    {
+                        Console.WriteLine();
+                    }
```
Almantask, line 43: Does too much in one if- extract to a private method.

Almantask (changes requested):
Exception is the only thing that really bothers me. Please change that. The advice on if is a common mistake, but it's a small one.

PR #68, opened 2021-05-25
Homework 7

Tailor-AI:
Implemented 2 different logger (File and Console) and added to RecipeApp homework. 

Not sure how to catch all unhandled exceptions but searched and found info about unhandled exception event which I subscribe to before start of RecipeApp form.


Tailor-AI:
I noticed that I have a catch (Exception ex) if writing to file fails. Will this become infinite loop when it then tries to write that unhandled exception to file again? Perhaps I should change this to only output text message.

Almantask (changes requested):
Please retarget the PR to lesson 4 so that you don't include my own code.

Tailor-AI:
Ok. I changed the base to lesson 4 if this is correct?

Almantask:
> Ok. I changed the base to lesson 4 if this is correct?

Yup, that is correct.

Almantask (approved):
Yup, looks good

PR #69, opened 2021-05-28
Homework/7 - Logging

cjahermanns:


Almantask (changes requested):
Please target lesson 4 as the PR branch, because I don't want to see my own code when reviewing changes.

PR #70, opened 2021-05-30
Homework/8/

cjahermanns:


ElectricityBillApi/Controllers/ElectricityProviderController.cs
```diff
@@ -0,0 +1,26 @@
...
+        public ElectricityProviderController(IElectricProviderPickerService electricProviderPickerService)
+        {
+            _electricProviderPickerService = electricProviderPickerService;
+        }
+
+        [HttpGet]
+        public IActionResult Get([FromBody] Address address)
+        {
+            var provider = _electricProviderPickerService.FindCheapest(address);
+            return Ok(new { Name = provider.Name, Price = provider.CalculatePrice(address)});
```
Almantask: I would move this: `Price = provider.CalculatePrice(address)}` to a different line, because it's a bit difficult to read. A line, just like a function- should do one ting.
In your case, it actually does 3 things:
returns Ok with a new object and calculates price.

Also, consider moving adding `FindCheapestProviderPrice` method to return strongly-typed cheapest provider and its price.

ElectricityBillApi/Models/ElectricProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    public class ElectricProvider
+    {
+        private PowerPlant _subscibedPowerPlant;
+
+        public string Name { get; set; }
+
+        public decimal CalculatePrice(Address address)
+        {
+            if (_subscibedPowerPlant == default) throw new NotImplementedException("What to do here? More infos needed!");
```
Almantask: Oh, but it is implemented. Instead, thow `ArgumentNullException` and provide the null argument name using `nameof`.

ElectricityBillApi/Models/ElectricProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public void Subscribe(PowerPlant powerPlant)
+        {
+            if (_subscibedPowerPlant is null)
+            {
+                _subscibedPowerPlant = powerPlant;
+                return;
+            }
+
+            throw new InvalidOperationException("Povider already subscribed to a power plant.");
```
Almantask: It's nice to know which provider is already subbed. So include the name of a provider in an exception.

ElectricityBillApi/Models/ElectricProviders.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System.Collections.Generic;
+
+namespace ElectricityBillApi.Models
+{
+    public class ElectricProviders : List<ElectricProvider>
```
Almantask: Avoid inheriting base data structures. You're not extending the capabilities of a list, you are implementing business requirements for electric providers. List is an implementation detail. Therefore, instead add your own interface and hide the fact that the underlying type is a list.

Almantask (changes requested):

Almantask:
Overall looks neat, but that's exactly why I would like you to pay attention to style remarks and change stuff accordingly.

ElectricityBillApi/Models/ElectricProvider.cs
cjahermanns: Ok, done.., My initial thought was to not send out potential sensitive information

ElectricityBillApi/Models/ElectricProviders.cs
cjahermanns: You mean in that case I should rather go with something like a `Providers` class with a field `List<ElectricProvider>` ?

ElectricityBillApi/Models/ElectricProvider.cs
cjahermanns: Hmm, in my mind I thought I don't want exceptions in a working program, so I should cover all possible flaws with good buisness logic...

ElectricityBillApi/Controllers/ElectricityProviderController.cs
cjahermanns: Changed it to return a `Tuple(ElectricProvider, decimal)` with the price calculated (since I already calculated it anyways to find the provider)

cjahermanns:
Sorry for taking so long... Just moved and have some other things to do rn :D

Almantask (approved):

PR #71, opened 2021-06-18
Homework 9

cjahermanns:
Merged Homework 8 changes then added some tests. Had to change the price calculation to be way better (easier).


ElectricityBillApi.Tests/ElectricProviderTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        public ElectricProviderTests()
+        {
+            _sut = new ElectricProvider();
+        }
+
+        [Fact]
+        public void CalculatePrice_CalculatesCorrectPrice()
+        {
+            // arrange
+            var loc1 = new Location()
```
Almantask: Is there any reason to have a location with just an x? Or does it logically always come with x, y and z? If the answer is yes, consider refactoring making the 3 coordinates readonly and be required by a contract (ctor)

ElectricityBillApi.Tests/ElectricProviderTests.cs
```diff
@@ -0,0 +1,94 @@
...
+                Y = 0,
+                Z = 0
+            };
+            var loc2 = new Location()
+            {
+                X = 1,
+                Y = 1,
+                Z = 1
+            };
+            decimal plantPrice = 5;
```
Almantask: You're buying electricity, not a plant.

ElectricityBillApi.Tests/ElectricProviderTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            var plant = new PowerPlant()
+            {
+                Location = loc1,
+                ElectricityPrice = plantPrice
+            };
+
+            // act
+            _sut.Subscribe(plant);
+            var actual = _sut.CalculatePrice(new Address() { Location = loc2 });
+            var expected = 3 * plantPrice;
```
Almantask: Why `*3`?

ElectricityBillApi.Tests/ElectricProviderTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                X = 1,
+                Y = 1,
+                Z = 1
+            };
+
+            // act
+
+            // assert
+            Assert.ThrowsAny<ArgumentNullException>(() => _sut.CalculatePrice(new Address() { Location = loc2 }));
```
Almantask: This is fine, but in the last lesson we learned about delegates. You can use instead:

```cs
// Act
Action calculatePriceNoPlantSubscribed = () => _sut.CalculatePrice(new Address() { Location = loc2 })

// Assert
Assert.ThrowsAny<ArgumentNullException>(calculatePriceNoPlantSubscribed );
```

Almantask: Though not entirely correct (the act happens to be called in an assert) this still segregates the assertion and the act nicely.

ElectricityBillApi.Tests/ElectricProviderTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        [Fact]
+        public void Subscribe_SubscribeTwoTimesThrows()
```
Almantask: try to use supplementary words to help identify the different parts of tests. Also segregate the different parts through `_`.
For example:
`Subscribe_WhenAlreadSubscribed_ThrowsInvalidOperationException`

Also specify the exception you throw for clarity.

ElectricityBillApi/Models/ElectricProvider.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+
+namespace ElectricityBillApi.Models
+{
+    public class ElectricProvider
+    {
+        private PowerPlant _subscibedPowerPlant;
+
+        public string Name { get; set; }
```
Almantask, line 9: Readonly probably is more suitable.

ElectricityBillApi/Models/ElectricProvider.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+
+namespace ElectricityBillApi.Models
+{
+    public class ElectricProvider
+    {
+        private PowerPlant _subscibedPowerPlant;
```
Almantask, line 7: readonly probably is more suitable.

ElectricityBillApi.Tests/ElectricProviderPickerServiceTests.cs
```diff
@@ -0,0 +1,166 @@
...
+            Assert.Equal(providers.First(), cheapestProvider);
+        }
+
+        [Fact]
+        public void FindCheapestProviderAndPrice_returnsFarCheapestProvider()
+        {
+            // arrange
+            var providers = GetTestData().ToList();
+            var iProviders = Substitute.For<IElectricProviders>();
+            iProviders.GetProviders.ReturnsForAnyArgs(providers);
```
Almantask: This is repeated multiple times, can be moved to a ctor for a shared setup. this will not cause a shared state, jut a shared logic for initialization.

ElectricityBillApi.Tests/ElectricProviderPickerServiceTests.cs
```diff
@@ -0,0 +1,166 @@
...
+                    Y=0,
+                    Z=0
+                }
+            };
+
+            // act 
+            var (cheapestProvider, calculatedPrice) = sut.FindCheapestProviderAndPrice(ownAddress);
+
+            // assert
+            Assert.Equal(providers.First(), cheapestProvider);
```
Almantask: If the order matters, don't use `IEnumerable`. Use an ordered list or an array or similar. When using `IEnumerable` the order of elements doesn't have to be kept.

Almantask (changes requested):

ElectricityBillApi.Tests/ElectricProviderTests.cs
cjahermanns: Good point! Added the ctor

ElectricityBillApi.Tests/ElectricProviderTests.cs
cjahermanns: Fixed the typo =)

ElectricityBillApi.Tests/ElectricProviderTests.cs
cjahermanns: Never really thought about using delegates here :D tyvm, will definitely use them more here!

ElectricityBillApi.Tests/ElectricProviderTests.cs
cjahermanns: Added a method to calc the expected price. Makes things more clear I hope.

ElectricityBillApi.Tests/ElectricProviderTests.cs
cjahermanns: Renamed them.

ElectricityBillApi.Tests/ElectricProviderPickerServiceTests.cs
cjahermanns: That's true, I use fields now.

ElectricityBillApi.Tests/ElectricProviderPickerServiceTests.cs
cjahermanns: Changed it to a regular `List<ElectricProvider>`.

ElectricityBillApi/Models/ElectricProvider.cs
cjahermanns: I assume you mean the field (like in the next request)...

ElectricityBillApi/Models/ElectricProvider.cs
cjahermanns: If I make the field readonly I cannot use my subscription logic...

ElectricityBillApi.Tests/ElectricProviderTests.cs
```diff
@@ -90,5 +77,14 @@ public void Unsubscribe_SubscribeTwoTimesDoesntThrow()
...
             _sut.Unsubscribe(plant);
         }
+
+        private double CalcLocationFactor(Location plantLocation, Location consumerLocation)
+        {
+            var xFac = Math.Abs(plantLocation.X - consumerLocation.X);
+            var yFac = Math.Abs(plantLocation.Y - consumerLocation.Y);
+            var zFac = Math.Abs(plantLocation.Z - consumerLocation.Z);
+
+            return xFac + yFac + zFac;
```
Almantask, line 87: Calculating distance fits a location class better than

ElectricityBillApi.Tests/ElectricProviderPickerServiceTests.cs
```diff
@@ -157,10 +134,13 @@ private IEnumerable<ElectricProvider> GetTestData()
...
-            yield return providerCheapFar;
-            yield return providerExpensiveNear;
-            yield return providerExpensiveFar;
+            return new List<ElectricProvider>() 
+            {
+                providerCheapNear,
+                providerCheapFar,
+                providerExpensiveNear,
+                providerExpensiveFar
+            };
```
Almantask, line 143: Yield return was the right way to do this. The downside of doing what you did now is that you always have 4 providers in memory. `yield return` is lazy loading- you only ever have 1 provider at a time and when the next case runs- the previous provider will be forgotten.

Almantask (approved):

PR #72, opened 2021-06-18
Homework/10

cjahermanns:
Is this what you meant with the assignment?

### DIFF
--- a/Homework2.Tests/ArrayExtensionsTests.cs
+++ b/Homework2.Tests/ArrayExtensionsTests.cs
@@ -34,19 +34,23 @@ namespace Homework2.Tests
         }
 
         [Theory]
-        [InlineData(4, 5, 6, 1, 3, 1, 3, 4, 5, 6)]
-        [InlineData(0, 1, 2, 3, 9, -5, -5, 0, 1, 2, 3, 9)]
-        [InlineData(9.1, 8.5, 1.0, -5.2, -5.2, 1.0, 8.5, 9.1)]
-        public void Sort_sorts_the_array(params object[] objs)
+        [InlineData(new int[] { 4, 5, 6, 1, 3 }, new int[] { 1, 3, 4, 5, 6 })]
+        [InlineData(new int[] { 0, 1, 2, 3, 9, -5 }, new int[] { -5, 0, 1, 2, 3, 9 })]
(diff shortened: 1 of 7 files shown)


Homework2.Tests/ArrayExtensionsTests.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    public class ArrayExtensionsTests
+    {
+        [Theory]
+        [InlineData(999, 4)]
+        [InlineData(999, 0)]
+        [InlineData(999, 10)]
+        public void AddAt_adds_at_idx(int item, int idx)
+        {
+            var testarray = new int[] { 11, 22, 33, 44, 55, 66, 77, 88, 99, 111 };
```
Almantask: Overkill test data. Having 3 elements would be enough.

Homework2.Tests/ArrayExtensionsTests.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Homework2.Tests
+{
+    public class ArrayExtensionsTests
+    {
+        [Theory]
+        [InlineData(999, 4)]
+        [InlineData(999, 0)]
+        [InlineData(999, 10)]
+        public void AddAt_adds_at_idx(int item, int idx)
```
Almantask, line 19: What happens if idx is higher than array size?

Almantask: What about negative?

Almantask: same for all methods here.

Almantask (changes requested):

Homework2.Tests/ArrayExtensionsTests.cs
cjahermanns: Made it smaller and used a ctor to init it.

Homework2.Tests/ArrayExtensionsTests.cs
cjahermanns: Changed existing and added new tests. Also added exceptions to the methods (and changed the `RemoveAt(int idx)` one).

Almantask (approved):

PR #73, opened 2021-07-02
Homework 8

Tailor-AI:
My first try for homework 8. 

Only used ElectricityProviderPicker in the controller so this was the only dependency I injected or should you inject the other class like ElectricityProvider aswell even if not used?

I used your Range class from the lesson with a twist to calculate the cost for end customer.


ElectricBillApi/Controllers/ElectricProviderController.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                var provider = _electricProviderPicker.FindCheapest(address);
+
+                return Ok($"{provider.Name}, {provider.CalculatePrice(address)}, {provider.CurrentSubscribedPlant}");
+            }
+            catch (NoSubscribedPowerPlantException ex)
+            {
+                return NotFound(ex.Message);
+            }
+            catch (NoValidAdressException ex)
```
Almantask: If address involves validation, it's better to just allow 500 to be returned- it happened in the server and it actually did some work.
400 implies that we didn't even start processing the request- the shape of request was not understood.

ElectricBillApi/Controllers/ElectricProviderController.cs
```diff
@@ -0,0 +1,77 @@
...
+                return BadRequest(ex.Message);
+            }
+            catch (NoValidPowerPlantException ex)
+            {
+                return BadRequest(ex.Message);
+            }
+            catch (NoProviderFoundException ex)
+            {
+                return BadRequest(ex.Message);
+            }
```
Almantask: no need to handle these- just let them fail naturally.

ElectricBillApi/Services/ElectricityProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+        public void Unsubscribe(PowerPlant plant)
+        {
+            if(currentSubscribedPowerPlant == null || currentSubscribedPowerPlant.Name != plant.Name)
+            {
+                return;
+            }
+            else
+            {
+                currentSubscribedPowerPlant = null;
+            }
```
Almantask: no need else, because you return.

ElectricBillApi/Services/ElectricityProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        private static double CalculateDistance (Location a, Location b)
+        {
+            return Math.Sqrt(Math.Pow((a.X - b.X), 2) + Math.Pow((a.Y - b.Y), 2) + Math.Pow((a.Z - b.Z), 2));
+        }
```
Almantask: Distance calculation fits a location class better.

Almantask (changes requested):

ElectricBillApi/Controllers/ElectricProviderController.cs
Tailor-AI: Ok. I removed the exception and returned 500 if provider is null.

ElectricBillApi/Controllers/ElectricProviderController.cs
Tailor-AI: Ok. Removed the exceptions and the implementations.

ElectricBillApi/Services/ElectricityProvider.cs
Tailor-AI: Fixed. Thx.

ElectricBillApi/Services/ElectricityProvider.cs
Tailor-AI: Yes, I have moved this to the Location class.

ElectricBillApi/Services/MyExtensions.cs
```diff
@@ -35,7 +19,8 @@ public static ElectricityProvider StringToProvider(this string name)
                     return provider;
                 }
             }
-            throw new NoProviderFoundException("No such electricityprovider exists.");
```
Almantask: Throwing was good. You just don't need to handle it in a controller, let it fail. It will fail with 500 naturally.

Almantask (changes requested):

ElectricBillApi/Services/MyExtensions.cs
Tailor-AI: Ok. Hopefully I understood it correctly now? However I dont get the 500 in Postman because VS will stop at the exception but it will be 500 when running compiled/standalone?

Almantask (approved):

PR #74, opened 2021-08-06
Homework/1-2-3

Zenixas:


Zenixas:
sorry for pushing all 3 in one, I'm not sure how I do seperate pull requests

Homework1/Homework1/Program.cs
```diff
@@ -0,0 +1,49 @@
...
+            double.TryParse(Console.ReadLine(), out double weight);
+
+            Console.WriteLine("What is your height?");
+            double.TryParse(Console.ReadLine(), out double height);
+
+            Print_Information(name, surname, age, weight, height);
+        }
+
+
+        static void Main(string[] args)
```
Almantask: This reads weird. Code should read like book. Entry functions should be at the top and the dependencies below them. In your case it's vice versa.

Homework2/Homework2/Program.cs
```diff
@@ -0,0 +1,183 @@
...
+                if (index != position && temp != NewArray.Length)
+                {
+                    NewArray[temp++] = array[index];
+                }
+            }
+
+            Console.WriteLine("\nBefore: ");
+            PrintArray(array);
+            Console.WriteLine("\nAfter: ");
+            PrintArray(NewArray);
```
Almantask: This console writeline is a side effect. Unless debugging, don't write lines like this

Almantask: If you do want to, write it outside of a function

Homework2/Homework2/Program.cs
```diff
@@ -0,0 +1,183 @@
...
+            int.TryParse(Console.ReadLine(), out int position);
+            RemoveAtAny(array, position);
+
+            Console.WriteLine("\n=== SIGN UP ===");
+            SignUp();
+
+            Console.WriteLine("\n=== LOGIN ===");
+            Login();
+
+            Console.ReadKey();
```
Almantask: Split to smaller functions

Almantask (changes requested):

Zenixas:
updated hw 1 and 2 issues

PR #75, opened 2021-08-11
solved chapter1 - exercise 1 

ghost:
- made class BMI Calculator
- made method bmi-calculator
- made method calculate


Chapter1/Array_Functions.cs
```diff
@@ -0,0 +1,105 @@
+﻿using System;
+
+namespace BootCamp.Chapter1
+{
+    public static class Array_Functions
+    {
+        public static int[] Sort_Array(int[] array)
+        {
+            int temp;
+            for (int j = 0; j <= array.Length - 2; j++)
```
Almantask: Why do both I and j have the same end? It means that you compare the same number twice.

Chapter1/BMI_Calculator.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter1
+{
+
+    public static class BMICalculator
```
Almantask, line 8: Shouldn't have been included in hw

Almantask (dismissed):

Chapter1/Array_Functions.cs
ghost: Lets say we have the array [ 4 2 3 5].

I goes then from 4 till the 5 
J goes then from 2 tilll the 5 
and then from 3 5 
and then only the 5 

So all numbers after a certain number are compared to that number.

Chapter1/BMI_Calculator.cs
ghost: We do not need the class or the static part ?

ghost:
Moment, this is the wrong code. This is exercise 2 or 3 not exercise 1
Schould I make a new PR for exercise 1 or make a new commit on this PR

Chapter1/BMI_Calculator.cs
```diff
@@ -0,0 +1,59 @@
...
+            var weight = Console.ReadLine();
+            Console.Write("Can I have your height in cm: ");
+            var height = Console.ReadLine();
+
+            Console.WriteLine($"{name} {surname} is {age} years old, his weight is {weight} kg and his height is {height} cm.");
+
+            var BMI = calculate(height, weight);
+            Console.WriteLine($"This person bmi is {BMI}");
+
+            // Person 2
```
aduckinpants: While this works I feel like it might be against the spirit of the exercise. Can you find a way to do the same without duplicating your code?

Chapter1/BMI_Calculator.cs
aduckinpants: I think this is a V1 Chap1 homework not V2?

Chapter1/BMI_Calculator.cs
```diff
@@ -0,0 +1,59 @@
...
+            BMI = calculate(height, weight);
+            Console.WriteLine($"This person bmi is {BMI}");
+
+
+
+        }
+
+        private static double calculate(string height, string weight)
+        {
+            var parsed_height = double.Parse(height) / 100;
```
aduckinpants: snake_case is not idiomatic in modern c#

Chapter1/BMI_Calculator.cs
```diff
@@ -0,0 +1,59 @@
...
+            Console.WriteLine($"{name2} {surname2} is {age2} years old, his weight is {weight2} kg and his height is {height2} cm.");
+
+            BMI = calculate(height, weight);
+            Console.WriteLine($"This person bmi is {BMI}");
+
+
+
+        }
+
+        private static double calculate(string height, string weight)
```
aduckinpants: Methods should be start UpperCase

Chapter1/Array_Functions.cs
```diff
@@ -0,0 +1,105 @@
+﻿using System;
+
+namespace BootCamp.Chapter1
+{
+    public static class Array_Functions
```
aduckinpants: See prev comment about snake_case.

aduckinpants (changes requested):

Chapter1/BMI_Calculator.cs
ghost: yep, I can use a for loop. Did not know I was allowed to use loops

Chapter1/BMI_Calculator.cs
ghost: this is v2 homework. See https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/wiki/C%23-Keywords-and-User-Input

oke, so better is parsedHeight ?

Chapter1/BMI_Calculator.cs
ghost: will change it

Chapter1/BMI_Calculator.cs
ghost: nope. this is v2 homework. See https://github.com/csinn/CSharp-From-Zero-To-Hero-v2/wiki/C%23-Keywords-and-User-Input

ghost: the array_functions is not the right one. that is challenge 3

aduckinpants: Gotcha

Chapter1/BMI_Calculator.cs
aduckinpants: A good rule of thumb is to follow the MS guidelines https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names

Chapter1/BMI_Calculator.cs
ghost: or do i rather see a seperate function/method for it

Chapter1/Array_Functions.cs
ghost: This is homework 2

Chapter1/BMI_Calculator.cs
ghost: changed

Chapter1/Array_Functions.cs
ghost: changed

Almantask (approved):

PR #79, opened 2021-08-14
solved chapter1 - exercise 2

ghost:
Did all these steps:

Add element at the start of an array
    Add element at the end of an array
    Add element at any position of an array
    Remove element at the start of an array
    Remove element at the end of an array
    Remove element at a given index of an array
    Login: takes username and password. If they match a hidden username and password- a message "Logged in!" is printed. Username capitalization does not matter, password capitalization must match.


Chapter1/Array_Functions.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            return Remove_Nth_Place_Array(array, array.Length -1 ); 
+        }
+
+        public static int[] Remove_Nth_Place_Array(int[] array, int index)
+        {
+            var new_array = new int[array.Length - 1];  
+            for (int i = 0; i < array.Length; i++)
+            {
+                if (i == index)
```
aduckinpants, line 79: Would the same trick used in `Add_Nth_Place_Array` allow you to reduce the branch comparisons in the loop?

aduckinpants (approved):

Chapter1/Array_Functions.cs
ghost: yep I could . Will try it

Chapter1/Array_Functions.cs
```diff
@@ -0,0 +1,105 @@
+﻿using System;
+
+namespace BootCamp.Chapter1
+{
+    public static class Array_Functions
+    {
+        public static int[] Sort_Array(int[] array)
```
Almantask, line 7: In C#, don't use snake_case.

Almantask: Use PascalCase

Almantask (approved):
Regardless of bad naming conventions, looks good. Not sure about the value of comments too, they seem to be explaining obvious stuff.

PR #80, opened 2021-08-17
Solved Chapter1 - Exercise 3 

ghost:


Chapter1_Challenge3/Authentication.cs
```diff
@@ -0,0 +1,110 @@
...
+            try
+            {
+                using (var stream = new StreamReader(file))
+                {
+                    contents = stream.ReadToEnd();
+                }
+            }
+            catch
+            {
+                throw new UsersNotFoundException();
```
aduckinpants: Per the spec

> If there is no Users.txt file- the application should throw "UsersNotFoundException.

This will throw the error for any exception where the spec instructs that it only applies when the file is not found exception is thrown.

Chapter1_Challenge3/Authentication.cs
```diff
@@ -0,0 +1,110 @@
...
+        
+
+        public static void Login()
+        {
+            Console.Write("May I have your username: ");
+            var username = Console.ReadLine();
+            Console.Write("May I have your password: ");
+            var password = Console.ReadLine(); 
+                        
+            var content = ReadFile();
```
aduckinpants, line 15: Food for thought here - In a real world type application it would make more sense to do this before gathering the input as there's not much point in continuing otherwise.

Chapter1_Challenge3/Authentication.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            foreach (var line in contentLines)
+            {
+                var user = line.Split(" ");
+                userdata[user[0]] = user[1].Trim('\r', '\n');    
+            }
+
+            foreach (var user in userdata)
+            {
+                if (user.Key == username && user.Value == password)
```
aduckinpants, line 28: Is the case of the operands important here? myPass v MyPass ?

Chapter1_Challenge3/Authentication.cs
```diff
@@ -0,0 +1,110 @@
...
+            try
+            {
+                using (var stream = new StreamWriter(file, false))
+                {
+                    stream.Write($"{content}{Environment.NewLine}{username} {password}");
+                }
+            }
+            catch
+            {
+                throw new UsersNotFoundException();
```
aduckinpants, line 98: See comment on this exception from prev PR

Chapter1_Challenge3/DuplicateUserCredentialsException.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Runtime.Serialization;
+
+namespace Chapter1_Challenge3
+{
+    [Serializable]
+    internal class DuplicateUserCredentialsException : Exception
+    {
+        public DuplicateUserCredentialsException()
+        {
+            Console.WriteLine("The user is already in the database");
```
aduckinpants: Which user is duplicated? If this is something meant for logging you'd want that sort of information available.

Chapter1_Challenge3/Menu.cs
```diff
@@ -0,0 +1,62 @@
...
+            Console.WriteLine("3) Exit programm");
+            Console.WriteLine();
+            Console.Write("Your Choice: "); 
+        }
+
+        public static int InputAndValidateChoice()
+        {
+           var valid = int.TryParse(Console.ReadLine(), out int choice);
+
+            if (!valid && choice < 1 || choice >3)
```
aduckinpants: This predicate is kind of awkward and hard to read, consider reworking the logic or adding parenthesis to make it more clear to the reader.

aduckinpants (changes requested):

Chapter1_Challenge3/Authentication.cs
ghost: oke, so I have to catch a specific exception? 
Then I have to figure out which one

Chapter1_Challenge3/Authentication.cs
ghost: I can change this so that the password and username are arguments when that is more a real world senario

Chapter1_Challenge3/Authentication.cs
ghost: MyPass is a class  so I do not understand your remark.

Chapter1_Challenge3/DuplicateUserCredentialsException.cs
ghost: oke, I have to find out how to pass the user to the exception then.

Chapter1_Challenge3/Menu.cs
ghost: maybe `(!valid && (choice < 1 || choice > 3))` ?

Chapter1_Challenge3/Authentication.cs
```diff
@@ -0,0 +1,110 @@
...
+            }
+
+            return contents; 
+        }
+
+        
+
+        public static void Login()
```
Almantask: I feel ike you could split this function into smaller ones to make it more clear what itcdoes

Chapter1_Challenge3/Authentication.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Text;
+
+namespace Chapter1_Challenge3
+{
+    public static class Authentication
+    {
+        const string file = @"user.txt";
+
+        public static string ReadFile()
+        {
```
Almantask, line 13: Does it have to be public?

Almantask: Since it is used by other functions, move it to the bottom.

Chapter1_Challenge3/Menu.cs
Almantask: Consider using constants to make it clear what is 1, 2 and 3...

Almantask (changes requested):

Chapter1_Challenge3/Authentication.cs
ghost: Solved by ony catching the filenotfoundException.

Chapter1_Challenge3/DuplicateUserCredentialsException.cs
ghost: Also solved I hope

Chapter1_Challenge3/Authentication.cs
ghost: I deleted the input part. so I do not know if I still need to split this one

Chapter1_Challenge3/Authentication.cs
ghost: no, I will make it private and copy it to the end of the file

Chapter1_Challenge3/Menu.cs
ghost: I tried it but because it's a static class I cannot add instance members.

Chapter1_Challenge3/Authentication.cs
Almantask: If you see any opportunities for splitting- do it. If not, it's good as-is.

Chapter1_Challenge3/Menu.cs
Almantask: Constants are static.
`const double PI = 3.14`  is in a static `Math` class.

Almantask: Still as bad. With constants it might be better.

Almantask (changes requested):

RoelofWobben:
yep, 

but as I wrote in a earlier respons. 
When I do  var LowestMenuChoice = 1 after the `static class Menu` 

I see this message : 
Menu.LowestMenuChoice': cannot declare instance members in a static class.

Or do I have to do that in the showMenu method ? 

And I tried but do not see a way to make the login method smaller

Chapter1_Challenge3/Authentication.cs
ghost: Sorry, I do not see it

Chapter1_Challenge3/Menu.cs
ghost: Must I then make the class not static because otherwise I see this error message 

Menu.LowestMenuChoice': cannot declare instance members in a static class.

or is there another way to make this work

ghost:
is this one approved or do I have to change anything ?

PR #82, opened 2021-08-21
Homework 7

Zenixas:
Fully added both console and file loggers. Easy ability to change from one to another inside ActiveLogger.cs file.


RecipeApp/RecipeApp.Core/Exceptions/Exceptions.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using RecipeApp.Core.Services.Logging;
+
+namespace RecipeApp.Core
+{
+    public class InvalidRecipeException : Exception
+    {
+        public InvalidRecipeException(string reason) : base(reason)
+        {
+            ActiveLogger.iLogger.Log(reason, LoggingLevels.levels.Warn);
```
Almantask: Should be.Instance. ILogger is not pascal case and besides by saying logger twice (in class and field name) you are redundant.

RecipeApp/RecipeApp.Core/Services/Logging/ILogger.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace RecipeApp.Core.Services.Logging
+{
+    public interface ILogger
+    {
+        public abstract void Log(string message, LoggingLevels.levels level);
```
Almantask: Why abstract?

RecipeApp/RecipeApp.Core/Services/Logging/Types/FileLogging.cs
```diff
@@ -0,0 +1,34 @@
...
+            {
+                writer.Write($"{logs}\n{log}");
+            }
+        }
+
+        private string Read()
+        {
+            using (StreamReader reader = new StreamReader(LogsFile))
+            {
+                return reader.ReadToEnd();
```
Almantask: Streams are usually fro more advanced cases. Use FilemWirteAllText, ReadAllText instead

RecipeApp/RecipeApp.Core/Services/RecipeConversion/ConvertToSiUnitsClass.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System.Globalization;
+
+namespace RecipeApp.Core
+{
+    internal class ConvertToSiUnitsClass : RecipeConverter
```
Almantask: Dotn use class in name

Almantask (changes requested):

RecipeApp/RecipeApp.Core/Services/RecipeConversion/ConvertToSiUnitsClass.cs
Zenixas: Understood, I had method with same exact name, so added word class to distinguish them. But I'll change them

RecipeApp/RecipeApp.Core/Services/Logging/ILogger.cs
Zenixas: I thought that it needed to be abstract, because I would need to create either console logger or file logger. But I might have misunderstood what it means.

Zenixas:
- Renamed the files and classes containing "Class" in their name.
- Renamed "ActiveLogger" class to "Instance" to avoid redundancies.
- Replaced streams with File.WriteAllText and File.ReadAllText methods
- Removed the unnecessary Abstract keyword.

